PR #10, opened 2017-09-20
Generate cities JSON

micapam:
This PR provides a script to generate JSON from CSV. 

Eventually we may provide something more user-friendly (maybe with a GUI, and ability to upload) once we want actual users to do this. This is just a quick way that will work for us, for now.

The idea is that the output JSON will be loaded along with 'data-shell.json' (at JSON path `$.data.cities`); the two files will be merged at runtime to provide the data. That loading and merging will be done in a separate Jira task, SC-48; this task just creates the cities JSON out of the CSV.


davidgilbertson (approved):
Approved

PR #12, opened 2017-09-21
CI/CD config

micapam:
Configuration for CI server, including deployment to CloudFoundry upon new commits to `master` and `staging` branches.

Ref: https://circleci.com/docs/2.0/

PR #13, opened 2017-09-21
Load cities data from JSON and merge with static data

micapam:
- Data in `public/data/smartCitiesData.json` now just contains cities;
- Merged with static data at runtime.


micapam:
n.b. Tests will fail until #12 is approved, merged and this branch rebased on master.

scripts/convert.js
```diff
@@ -5,17 +5,19 @@
...
 const fs = require('fs');
 const mapValues = require('lodash/mapValues');
+const kebabCase = require('lodash/kebabCase');
 const getStdin = require('get-stdin');
 const path = require('path');
 
-const dataShellPath = path.resolve(__dirname, '../resources/data-shell.json');
-const dataShell = JSON.parse(fs.readFileSync(dataShellPath, 'utf-8'));
+const dataShellPath = path.resolve(__dirname, '../src/constants/data-shell.json');
+const dataShell = require(dataShellPath);
```
davidgilbertson: Unlike `readFile()`, `require()` is always relative to the current module so by convention should just be the relative path (the [full rules](https://nodejs.org/api/modules.html#modules_all_together) for resolving paths).

```js
const csv = require('csv');
const mapValues = require('lodash/mapValues');
const kebabCase = require('lodash/kebabCase');
const getStdin = require('get-stdin');
const dataShell = require('../src/constants/data-shell.json');

const dict = mapValues(dataShell.data.indices, (atts) => atts.source);

// etc.
```

scripts/convert.js
```diff
@@ -5,17 +5,19 @@
 const csv = require('csv');
 const fs = require('fs');
```
davidgilbertson: No longer used. There are other lint errors in this file too.

scripts/convert.js
```diff
@@ -5,17 +5,19 @@
...
 
 getStdin().then(rawCsv => {
   csv.parse(rawCsv, { columns: true }, (err, data) => {
     process.stdout.write(JSON.stringify(data.map((row) => ({
-      name: row['Cities'],
+      name: row.Cities,
+      id: kebabCase(row.Cities),
       indices: mapValues(dict, (colKey) => {
         if (!row[colKey]) throw new Error(`Key not found! ${colKey}`);
         return row[colKey];
```
davidgilbertson, line 20: I'm not convinced that the ability to pass in any input and output path is a good thing. How does a user of this script know where to output the file?

I'm a fan of developing for today, and right now the front end code relies on the data being output to a specific directory, but this code doesn't enforce outputting the data to that directory. So this is flexible, but if you specified something other than the `DATA_URL` constant, it fail silently since the new data would be ignored and the old data would still be used.

I think this should either use the `DATA_URL` URL in this file, or if we want to allow for it to have a different output at runtime, define a script in `package.json` that defines where the output must go:

`"generate-city-data": "node scripts/convert.js < resources/cities.csv > public/data/smartCitiesData.json"`

Since that documents the expected input/output too. Also we can then call this from the `build` script to ensure we always have the most recent data.

If we wind up having a different script to go to S3 or something, we update the script.

davidgilbertson (approved):
Approved, but not having the required paths hard-coded or documented seems brittle.

davidgilbertson (changes requested):
Sorry, I'm unapproving :) Lint errors should be fixed.

scripts/convert.js
micapam: Gotcha

scripts/convert.js
micapam: I like the idea of putting specifics in package.json. However, these suggested changes aren't in scope for this PR.

scripts/convert.js
```diff
@@ -3,19 +3,18 @@
...
+const kebabCase = require('lodash/kebabCase');
 const getStdin = require('get-stdin');
-const path = require('path');
+const dataShell = require('../src/constants/data-shell.json');
 
-const dataShellPath = path.resolve(__dirname, '../resources/data-shell.json');
-const dataShell = JSON.parse(fs.readFileSync(dataShellPath, 'utf-8'));
 const dict = mapValues(dataShell.data.indices, (atts) => atts.source);
 
 getStdin().then(rawCsv => {
```
davidgilbertson: This is still a lint error. Are you using a linter plugin for VS Code? It should pick up our `.eslintrc.js` file, but if not you might need to tell it to look there.

davidgilbertson (changes requested):
Still some lint errors

scripts/convert.js
micapam: Ahh, I was assuming CodeClimate would pick up eslint errors. I have now added the plugin to VS Code, I can see the errors.

davidgilbertson (approved):
LGTM

PR #14, opened 2017-09-22
Feat/sc 4 base styles

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-4

The style guide is at http://localhost:8080/style-guide

And looks like...
localhost_8080_style-guide 3


src/components/App/_globalStyles.scss
```diff
@@ -4,6 +4,14 @@ body {
   font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
   line-height: 1.6;
   margin: 0;
-  min-width: $app_min_width;
+  min-width: $app-min-width;
```
davidgilbertson, line 7: I put a note in the readme about naming conventions for CSS and updated underscores to dashes in variables for consistency.

src/components/StyleGuide/StyleGuide.js
```diff
@@ -0,0 +1,371 @@
+/* eslint-disable dot-notation */
+import React from 'react';
+import ColorRow from './ColorRow/ColorRow';
+import TypographyRow from './TypographyRow/TypographyRow';
+import style from './StyleGuide.scss';
+
+const COLORS = {
```
davidgilbertson: OK there is a ridiculous amount of duplication between mixins, class names, and content in this file, but it's only for the style guide and should not change.

src/components/StyleGuide/StyleGuide.scss
```diff
@@ -0,0 +1,93 @@
...
+  font-size: 5px; // to ensure font-sizes are set on each mixin
+  padding: 25px 25px 200px;
+  background: $color-white;
+}
+
+.colorsWrapper {
+  margin-bottom: 100px;
+}
+
+.color-navigation { background: $color-navigation; }
```
davidgilbertson: More duplication...

src/style/_typography.scss
```diff
@@ -0,0 +1,85 @@
+@mixin detail-text {
+  font-size: 16px;
+}
+
+@mixin detail-text--bold {
```
davidgilbertson: I suspect the letter spacing is going to be off on some of these, so they might get updated as we get to actually using these.

stylelint.config.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
   extends: 'stylelint-config-standard',
   ignoreFiles: '**/_normalize.scss',
+  rules: {
+    'at-rule-no-unknown': null,
```
davidgilbertson, line 5: Weirdly, the linter doesn't recognise `@mixin`, even when told it's dealing with Sass.

src/components/StyleGuide/StyleGuide.js
```diff
@@ -0,0 +1,355 @@
+/* eslint-disable dot-notation */
+import React from 'react';
+import ColorRow from './ColorRow/ColorRow';
+import TypographyRow from './TypographyRow/TypographyRow';
+import style from './StyleGuide.scss';
+
+const COLORS = [
```
micapam: Could we not DRY up this big array?

src/components/StyleGuide/StyleGuide.scss
```diff
@@ -0,0 +1,101 @@
...
+}
+
+.sectionText {
+  @include type-h1;
+}
+
+.subSectionText {
+  @include type-h3;
+}
+
```
micapam, line 20: Again, how about DRYing this up? @each loop over a $colors array?

micapam (approved):
Minor suggestions to DRY up some repetition, otherwise LGTM

src/components/StyleGuide/StyleGuide.scss
davidgilbertson: I gave it 15 minutes of trying then gave up. Using interpolation to get the variable names (in the `@each`) doesn't seem possible.

micapam:
Great - DRYing up not only reduces lines but also makes the special cases explicit and easy to find (in the code).

PR #18, opened 2017-09-26
Feat/sc 57 esm

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-57

This allows scripts that *aren't* parsed by Webpack/Babel to use ES6 modules too.


package.json
```diff
@@ -127,5 +129,8 @@
     "presets": [
       "react-app"
     ]
+  },
+  "@std/esm": {
+    "esm": "js"
```
davidgilbertson, line 134: I could not work out how to get Babel to *not* transpile modules. So rather than have `@std/esm` handle everything, it will only handle the convert script. Setting this to `js` means it will parse everything as an 'es6' module.

src/app.js
```diff
@@ -5,18 +5,23 @@ import { Provider } from 'react-redux';
...
   .then(checkStatus)
   .then(parseBody)
   .then((cities) => {
-    dataShell.data.cities = cities;
-    const store = configureStore(dataShell.data);
+    const store = configureStore({
+      cities,
+      indices: INDICATORS,
+      categories: CATEGORIES,
+    });
```
davidgilbertson, line 25: We don't need these in the store any more, but I didn't want store changes in this branch.

scripts/convert.js
```diff
@@ -1,14 +1,15 @@
 #!/usr/bin/env node
```
micapam: Shebang should be changed to: 
```
#!/usr/bin/env node -r @std/esm
```

micapam (changes requested):
Just needs the shebang fixing on the script; once that's fixed I'll approve. I like the separation of those different kind of data (categories, indices) rather than their being mushed together in data shell file. Also good that we don't have to rename everything :)

scripts/convert.js
davidgilbertson: Fixed

micapam (approved):

package.json
```diff
@@ -86,6 +87,7 @@
   "scripts": {
     "start": "node scripts/start.js",
     "build": "node scripts/build.js",
+    "convert": "node -r @std/esm scripts/convert.js < resources/cities.csv > public/data/smartCitiesData.json",
```
micapam: Oh BTW - I already added this - if you rebase off master you'll see. Best to remove this line

davidgilbertson: Fixed

PR #20, opened 2017-09-28
Feat/sc 5 header etc

micapam:
Site header and category nav. 

Caveat: colours are not 100% right yet on the menu options (active state should have different tint from the hover state). Also: not yet responsive.

Preview: <http://smart-cities-SC5.apps.staging.digital.gov.au>

Also in this PR: a bash script (`scripts/commit_pr.sh`) that commits, spins up an instance for QA, and opens a browser tab to automatically launch a PR with link to that instance. (This PR is the first created thus!)


package.json
```diff
@@ -36,6 +36,7 @@
     "react-motion": "^0.5.0",
     "react-redux": "5.0.6",
     "react-router-dom": "4.2.2",
+    "react-svg": "^2.1.21",
```
davidgilbertson: I know I'm not reviewing yet, but this can go. Check out ICT or my branch to see SVG icon usage.

micapam:
Scroll is working but the code is ugly. Now to refactor.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
```diff
@@ -1,34 +1,61 @@
...
+
+        {this.props.categories.map(category => (
+          <NavLink
+            className={classnames(style.link,
+              style[`link__${camelCase(`category ${category.id}`)}`])}
+            activeClassName={style.link__active}
+            key={category.id}
+            to={`/${this.props.cityId}/${category.id}`}
+          >
+            <span ref={(el) => {
```
micapam: I'm not very happy with this. The problem I faced was that NavLink, as a stateless functional component, cannot set a reference via `ref` (it just sets a `activeLinkEl` to null). I need to have a reference because I can't calculate how far to scroll until everything is rendered - without duplicating the same calculations are being done by the DOM, which would be the worst possible solution. (See Nick's 'mobile nav' mockups in InVision for the kind of scrolling that needed to be implemented.)

So, I am nesting the label inside a `<span>`, simply to get a reference, and then checking the parent element in the `scrollToActiveNavLink` function. This is gross, but I can't see another way to do it. Suggestions welcome.

package-lock.json
```diff
@@ -10557,6 +11457,7 @@
         "anymatch": "1.3.2",
         "exec-sh": "0.2.1",
         "fb-watchman": "2.0.0",
+        "fsevents": "1.1.2",
```
davidgilbertson: I don't think this file should be changed in this branch, should it? Can probably be reverted.

src/components/AllCitiesCategoryContainer/AllCitiesCategory/AllCitiesCategory.scss
```diff
@@ -2,4 +2,5 @@
 
 .main {
   background: $color-white;
+  overflow-x: hidden; // Hack while testing responsive layout - TODO remove
```
davidgilbertson: TODO, remove (actually I deleted this file in my branch

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
```diff
@@ -1,34 +1,61 @@
...
+        <NavLink
+          className={classnames(style.link, style.link__overview)}
+          activeClassName={style.link__active}
+          to={`/${this.props.cityId}/${NO_CATEGORY}`}
+        >Overview</NavLink>
+
+        {this.props.categories.map(category => (
+          <NavLink
+            className={classnames(style.link,
+              style[`link__${camelCase(`category ${category.id}`)}`])}
```
davidgilbertson: This should use the same logic as the `<Card>`, which does the same thing (apply a colour based on the category). The categories have a property called `colorName` which you can use here. There might be some logic that can be shared between the two.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
```diff
@@ -1,12 +1,99 @@
...
+    padding-left: 30px;
+  }
+}
+
 .link {
   display: inline-block;
-  padding: 10px;
+  padding: 0 20px;
   text-decoration: none;
+  height: 23px;
```
davidgilbertson: It's hard to vertically centre text when setting the height.

image

By setting only padding you'll get better results, e.g. `padding: 9px 20px 12px;` (or whatever works) and remove the height property.

Also, for good UX links should have a minimum tap target of at least 40px (less than that will fail some accessibility tests). Nothing serious but I'd mention that to Nick.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
```diff
@@ -1,12 +1,99 @@
 @import '~style/variables';
 
+.categoryNav {
+  background-color: $color-white;
+  display: flex;
+  height: 48px;
```
davidgilbertson: This works, but when child items have height set too, setting padding (instead of height) on the parent element means changes to child element height just work.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
```diff
@@ -1,12 +1,99 @@
...
+}
+
 .link {
   display: inline-block;
-  padding: 10px;
+  padding: 0 20px;
   text-decoration: none;
+  height: 23px;
+  border-radius: 100px;
+  font-size: 12px;
```
davidgilbertson: We shouldn't need `font-size` in the CSS anywhere, there's a typography style for this (e.g. `@include type-nav-horizontal`).

You can add breakpoints into the typography styles, you'll see I've done this in my PR.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
```diff
@@ -1,34 +1,61 @@
...
+            activeClassName={style.link__active}
+            key={category.id}
+            to={`/${this.props.cityId}/${category.id}`}
+          >
+            <span ref={(el) => {
+              if (this.props.categoryId === category.id) this.activeLinkEl = el;
+            }}
+            > {category.name} </span>
```
davidgilbertson: Minor: better not to rely on spaces for spacing (prefer padding).

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
```diff
@@ -1,12 +1,99 @@
...
+  box-shadow: 0 0 4px 2px rgba(0, 0, 0, 0.1);
+
+  @include for-desktop-up {
+    font-size: 16px;
+    height: 27px;
+    box-shadow: initial;
+
+    &:hover {
+      color: $color-white;
+      box-shadow: 0 0 4px 2px rgba(0, 0, 0, 0.1);
```
davidgilbertson: Does it make sense to centralise this? You can do a mixin for shadows of different sizes for consistency. I'm not sure if we have any/many other shadows though.

src/components/Header/Header.scss
```diff
@@ -1,9 +1,150 @@
...
 
-.wrapper {
-  border: 1px solid blue;
+header {
+  display: flex;
+  flex-direction: column;
+  background-color: $color-primary--500;
+  color: $color-white;
+
+  > * {
```
davidgilbertson: Minor: *generally* selectors like this don't scale well. It becomes easy to accidentally apply rules in places you don't mean. My preferred approach is to only ever target by class name (i.e. no wildcards, no tag names). It's also a theoretical performance hit because the browser matches CSS selectors right to left. So will match all components first, then filter that down to "all components that have a parent element with a `header` class.

But I suspect there's no real world impact on a site this size, and as for selectors this is probably not a real thing to worry about either. Feel free to ignore.

src/components/Header/Header.scss
```diff
@@ -1,9 +1,150 @@
...
+    @include for-desktop-up {
+      padding-left: 30px;
+      padding-right: 30px;
+      max-height: 176px;
+    }
+  }
+
+  a {
+    color: white;
```
davidgilbertson: `$color-white`

src/components/Header/Header.scss
```diff
@@ -1,9 +1,150 @@
...
+    text-decoration: underline;
+    text-decoration-color: $color-highlight;
+  }
+}
+
+.govContainer {
+  display: flex;
+  height: 48px;
+  padding-left: 14px;
+  flex-flow: column wrap;
```
davidgilbertson, line 18: Minor: I think this CSS could be simplified quite a lot by having one more nesting level of `<div>` to arrange the header items into left/right. (IMO more complex HTML is better than more complex CSS)

Then it's a simple two-column setup (also this is how the ICT header is done). Each column doesn't even need to be flex since it just contains rows.

It'll remove a lot of fiddly things like `flex: 0 0 100%` and `align-content: flex-start;` and having a the funky column-wrap thing.

src/components/Header/Header.scss
```diff
@@ -1,9 +1,150 @@
...
+  line-height: 46px;
+  margin-top: 10px;
+  max-height: initial;
+}
+
+.compactTitle {
+  font-size: 14px;
+  font-weight: 600;
+  max-height: 16px;
+  flex: 0 0 16px;
```
davidgilbertson: Minor, it took me a long while to work out that these are all actually in a wrapping column, so `16px` is really height here. To be more explicit (if you keep the whole flex-column setup) I think you can have `height: 16px` and `flex: none`.

I think `max-height` can go, no?

src/components/Header/Header.scss
```diff
@@ -1,9 +1,150 @@
...
+  }
+}
+
+.govAu {
+  font-size: 14px;
+  font-weight: 500;
+}
+
+.govWebsite {
+  font-size: 12px;
```
davidgilbertson, line 59: I *think* font-size should go from everywhere. The dream is to have all typography defined in one place. This works well if there's a finite number of styles, and I think Nick is keen to have that. But if every bit of text is different maybe that's all a waste of time.

Thoughts?

src/constants/categories.js
```diff
@@ -16,7 +16,7 @@ export default [
   {
     id: 'infrastructure',
     colorName: COLOR_NAMES.INFRASTRUCTURE,
-    name: 'Infrastructure & Other Things',
+    name: 'Infrastructure',
```
davidgilbertson: I think this will actually need to be a `navName` property since the page title is still "Infrastructure & Other Things". Same as "Livability & sustainability" and "Liveability" for the tab name.

.codeclimate.yml
```diff
@@ -0,0 +1,28 @@
...
+    checks: 
+      import/extensions:
+        enabled: false
+  fixme:
+    enabled: true
+ratings:
+  paths:
+  - "**.inc"
+  - "**.js"
+  - "**.jsx"
```
davidgilbertson: `**.jsx` can go, no?

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
davidgilbertson: I'm not sure that this is better, but you can get away without using refs by just responding to an `onClick` event passed into `<NavLink>`. To scroll to the current one when the page loads you can do the pretty nasty thing of looking it up by class name. It's bad form, but I think if you set it as a property not much can go wrong.

```jsx

function scrollElementIntoView(el) {
  el.parentElement.scrollLeft = el.offsetLeft - 40; // eslint-disable-line no-param-reassign
}

class CategoryNav extends Component {
  constructor(props) {
    super(props);

    this.activeElementClassName = style.link__active;
  }

  componentDidMount() {
    const activeEl = document.querySelector(`.${this.activeElementClassName}`);
    scrollElementIntoView(activeEl);
  }

  render() {
    return (
      <div className={style.categoryNav}>
        <NavLink
          className={classnames(style.link, style.link__overview)}
          activeClassName={this.activeElementClassName}
          to={`/${this.props.cityId}/${NO_CATEGORY}`}
        >Overview</NavLink>

        {this.props.categories.map(category => (
          <NavLink
            key={category.id}
            className={classnames(style.link,
              style[`link__${camelCase(`category ${category.id}`)}`])}
            activeClassName={this.activeElementClassName}
            to={`/${this.props.cityId}/${category.id}`}
            onClick={(e) => {
              scrollElementIntoView(e.target);
            }}
          >
            {category.name}
          </NavLink>
        ))}
      </div>
    );
  }
}
```

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
```diff
@@ -1,34 +1,61 @@
...
-    ))}
-  </div>
-);
+  componentDidUpdate() {
+    this.scrollToActiveNavLink();
+  }
+
+  scrollToActiveNavLink() {
+    if (this.activeLinkEl) {
+      this.el.scrollLeft = this.activeLinkEl.parentElement.offsetLeft - 20;
```
davidgilbertson: IMO 40 looks better because you can see that there's more options off to the left.

davidgilbertson (changes requested):
Back to you.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
micapam: Ahhh I hadn't thought of pulling the class name out of `style`. I thought the fact CSS Modules modifies the class names would preclude that. I prefer your implementation; at least there's not that extra HTML element.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
micapam: Which PR, @davidgilbertson ?

micapam:
Some feedback from Nick: 

- extend tap area:
image

- text should be white on hover for the category items: 
image

- centre text vertically on nav links:
image

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
micapam: Deferring to Nick on this

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
micapam: This whitespace has no visual effect, it's just for readability

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
```diff
@@ -1,34 +1,62 @@
...
+            activeClassName={style.link__active}
+            key={category.id}
+            to={`/${this.props.cityId}/${category.id}`}
+          >
+            <div
+              className={style[category.colorName]}
+              ref={(el) => {
+                if (this.props.categoryId === category.id) this.activeCategoryEl = el;
+              }}
+            > {category.navName || category.name} </div>
```
davidgilbertson: Minor: My first thought when I see this is that there are spaces for a reason and they must be doing something. It seems funny to add whitespace in the markup on purpose knowing that it will be stripped out by React. It isn't done anywhere else (e.g. `<div className={style.overview}>Overview</div>` above).

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
```diff
@@ -1,34 +1,62 @@
...
-    ))}
-  </div>
-);
+  componentDidUpdate() {
+    this.scrollToActiveNavLink();
+  }
+
+  scrollToActiveNavLink() {
+    if (this.activeCategoryEl) {
+      this.el.scrollLeft = this.activeCategoryEl.parentElement.offsetLeft - 20;
```
davidgilbertson: This has a bug on iOS where it selects two items:

image

I think what happens is that the DOM updates while your finger is still on the screen. So I click on "Jobs & Skills" and it moves to the left, my finger is now over the "Housing" button, so iOS applies the hover/active style and for whatever reason after I lift up, it doesn't remove the hover style.

This fixes it:
```
if (this.activeCategoryEl) {
  setTimeout(() => {
    this.el.scrollLeft = this.activeCategoryEl.parentElement.offsetLeft - 20;
  });
}
```

When you do `setTimeout()` without the (optional) delay parameter it puts the callback function on the 'next tick'

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
```diff
@@ -1,12 +1,89 @@
...
+      box-shadow: initial;
+
+      &:hover {
+        box-shadow: 0 0 4px 2px rgba(0, 0, 0, 0.1);
+      }
+    }
+  }
+
+  &:hover,
+  &.link__active {
```
davidgilbertson, line 50: This doesn't have different colours for the hover and active states. We have two options:
* do this task so that it can show different colours on hover, and just do, say `--500` for hover and `--600` for active. The header is then functional and we just update the colour values when we have them.
* We extend the task that updates the colours to add in functionality here.

Doesn't really matter, but I think it's better to lock down functionality in the task that describes it.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
```diff
@@ -1,12 +1,89 @@
...
+    .overview {
+      border: 1px solid $color-grey--500;
+      background-color: $color-white;
```
davidgilbertson: `background-color: $color-white;` isn't needed here.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
```diff
@@ -1,12 +1,89 @@
...
+      color: $color-grey--700;
+    }
+
+    .jobs {
+      background-color: $color-jobs--600;
```
davidgilbertson: You don't need the longhand `background-color`, `background` is fine and what's used elsewhere.

.codeclimate.yml
```diff
@@ -0,0 +1,27 @@
...
+      languages:
+      - ruby
+      - javascript
+      - python
+      - php
+  eslint:
+    enabled: false
+    checks: 
+      import/extensions:
+        enabled: false
```
davidgilbertson, line 15: I have no idea how this works, but isn't this disabling twice? eslint, then import/extensions?

scripts/commit_pr.sh
```diff
@@ -0,0 +1,30 @@
...
+INSTANCE="http://$SUBDOMAIN.apps.staging.digital.gov.au"
+
+git add -A
+git commit -m "$COMMIT_MSG"
+git push 
+
+npm run build 
+cf push $SUBDOMAIN -c "null"
+
+open "https://github.com/govau/smart-cities/compare/$URL_BRANCH?expand=1&body=Preview:+<$INSTANCE>"
```
davidgilbertson, line 30: Is it worth having a note in the readme about what this is?

davidgilbertson (changes requested):
There's an iOS bug that needs some attention...

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
micapam: I've shifted the spaces to the other side of the curly braces :)

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
micapam: Option A seems good, then we are just adding colour values in later.

.codeclimate.yml
micapam: It is, but I left it in as a reminder that import/extensions check appears to be buggy, so even if we decide to re-enable eslint, we should keep that one switched off.

scripts/commit_pr.sh
micapam: I am actually going to refactor this script in a separate branch, at which point I'll document it ... I'm going to give it opts so it can suit a bunch of cases (not just the assumptions at top of script) and also provide teardown. 

This might end up as something sharable with other projects that have a similar workflow.

src/components/Header/Header.scss
```diff
@@ -1,9 +1,142 @@
...
+}
+
+.orgLogo {
+  height: 65px;
+  width: 102px;
+}
+
+.contactLink {
+  font-size: 22px;
+  align-self: flex-end;
```
davidgilbertson: My bad for not testing this in IE till now, I really should have done it all in one go... but it's broke:

image

There's a few things:
* I told you about `initial`, not realising it's [not supported in IE](https://developer.mozilla.org/en-US/docs/Web/CSS/initial). Fixing this (I'll put other comments inline) fixes the drop shadows showing and the top text not showing.
* The header being too tall is likely a flexbox bug. A clever chap has created a page that chronicles every bug with flexbox, worth having a read through: https://github.com/philipwalton/flexbugs.

Do you have a Windows VM? Do you know about modern.ie?

src/components/Header/Header.scss
```diff
@@ -1,9 +1,142 @@
...
+  background-color: $color-badge;
 }
 
 .title {
-  font-size: 20px;
+  font-size: 36px;
+  font-weight: 300;
+  line-height: 46px;
+  margin-top: 10px;
+  max-height: initial;
```
davidgilbertson: `initial` should go, is `max-height` needed at all - it's not overriding anything I can see.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
```diff
@@ -1,12 +1,116 @@
...
+  div {
+    @include type-nav-horizontal;
+
+    padding: 0 24px;
+    border-radius: 100px;
+    color: $color-grey--700;
+    box-shadow: 0 0 4px 2px rgba(0, 0, 0, 0.1);
+
+    @include for-desktop-up {
+      box-shadow: initial;
```
davidgilbertson: Should be `box-shadow: none;`. Sorry :(

src/components/Header/Header.scss
```diff
@@ -1,9 +1,142 @@
...
+    display: none;
+  }
+}
+
+.govAu,
+.govWebsite {
+  display: none;
+
+  @include for-desktop-up {
+    display: initial;
```
davidgilbertson: This is trickier. Technically `display: inline;` will work, but will break if those spans become anything but spans. Good enough for now I'd say.

davidgilbertson (changes requested):
Oh no, IE.

src/components/Header/Header.scss
micapam: I ended up doing this to satisfy IE.

src/components/Header/Header.scss
micapam: I can't see a global style that corresponds to this. I think: best to set global styles in typography file, but not clutter it with snowflakes.

src/components/Header/Header.scss
```diff
@@ -66,15 +64,22 @@ header {
...
     justify-content: space-between;
-    margin-bottom: 24px;
   }
 }
 
+.titleContainer,
+.orgContainer {
+  padding: 0 32px;
+  max-height: 176px;
+  display: inline-flex;
```
davidgilbertson, line 77: Just an FYI. Since the parent of these two is `display: flex`, these guys are 'flex items', so there is no difference between `inline-flex` and `flex` (so this could just be flex).

davidgilbertson (approved):
LGTM!

PR #23, opened 2017-10-03
Feat/sc 58 paths as entry point

micapam:
Fixes SC-58. Try this URL (and reload, open other URLs, etc) -

http://smart-cities-sc-58.apps.staging.digital.gov.au/all-cities/liveability


public/nginx.conf
```diff
@@ -0,0 +1,49 @@
...
+  server_tokens off;
+
+  server {
+    listen <%= ENV["PORT"] %>;
+    server_name localhost;
+
+    location / {
+      root <%= ENV["APP_ROOT"] %>/public;
+      index index.html index.htm Default.htm;
+      try_files $uri /index.html;
```
micapam, line 34: n.b. This line (with `try_files`) is the only really new line - that's the one that makes our SPA work as an SPA.

The rest is copied from the default nginx.conf provided in the version of staticfile-buildpack (1.2.2) used by Cloud team.

micapam:
Some feedback from Nick: 

- extend tap area:
image

- text should be white on hover for the category items: 
image

- centre text vertically on nav links:
image

davidgilbertson (approved):
LGTM

PR #24, opened 2017-10-04
Feat/sc 59 sub category summaries

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-59

The sub-category section is likely to change so the layout is not pixel-perfect. It should still work on all screen sizes though.

I've also added in placeholder sections for the chart so that I could build the functionality to jump to the chart section.

I've filled in a few indicators to facilitate testing but it's still pretty scrappy.


src/components/CategoryBanner/CategoryBanner.js
```diff
@@ -1,45 +1,16 @@
...
-function aggregateIndicatorForCities({ cities, indicator, aggregationMethod }) {
-  if (aggregationMethod === AGGREGATION_METHODS.AVERAGE) {
-    return sumIndicators(cities, indicator) / cities.length;
-  } else if (aggregationMethod === AGGREGATION_METHODS.SUM) {
-    return sumIndicators(cities, indicator);
-  }
-
-  console.error(`${aggregationMethod} is not a valid aggregation method`);
-  return null;
-}
```
davidgilbertson, line 26: This was all moved out into a separate file

src/components/SubCategorySummary/SubCategorySummary.js
```diff
@@ -0,0 +1,74 @@
...
+import PropTypes from 'prop-types';
+import IndicatorCard from '../Card/IndicatorCard/IndicatorCard';
+import aggregateIndicatorForCities from '../../helpers/aggregateIndicatorForCities';
+import getSubCategorySectionId from '../../helpers/getSubCategorySectionId';
+import { INDICATORS } from '../../constants';
+import style from './SubCategorySummary.scss';
+import COLORS from '../../style/_colors.scss';
+
+const SubCategorySummary = (props) => {
+  const backgroundColor = COLORS[`${props.colorName.toUpperCase()}_${props.colorVariation}`];
```
davidgilbertson, line 11: This could *maybe* be a separate function but it isn't used anywhere else and I think would just add complexity.

src/components/SubCategorySummary/SubCategorySummary.js
```diff
@@ -0,0 +1,74 @@
...
+          indicator={indicatorId}
+          value={aggregateIndicatorForCities(
+            indicatorId,
+            props.cities,
+          )}
+        />
+        {lastOne && (
+          <a
+            className={style.linkWrapper}
+            href={`#${getSubCategorySectionId(props.name)}`}
```
davidgilbertson, line 34: This is just a straight jump to the section, we can do a fancy slide later if we have the time

src/helpers/aggregateIndicatorForCities.js
```diff
@@ -0,0 +1,22 @@
...
+  AGGREGATION_METHODS,
+  INDICATORS,
+} from '../constants';
+
+export default function aggregateIndicatorForCities(indicatorId, cities) {
+  const values = cities.map(city => city.indices[indicatorId]);
+  const aggregationMethod = INDICATORS[indicatorId].aggregationMethod;
+
+  if (aggregationMethod === AGGREGATION_METHODS.MEDIAN) {
+    return median(values);
```
davidgilbertson, line 15: Borderline not worth a library just to do a median, but it's a tiny lib.

src/constants/indicators.js
```diff
@@ -1,157 +1,159 @@
...
     source: 'Population, All persons, People',
   },
   growthRate: {
     name: 'The growthRate',
-    shortDescription: 'The growthRate index is all about growthRate',
-    longDescription: 'The growthRate index is all about growthRate and event more!!',
-    cardPrefix: '',
+    shortDescription: 'TODO',
+    longDescription: 'TODO',
     cardSuffix: '/year',
```
micapam, line 52: This should be 'per year' rather than '/year', esp. if used with a percentage - otherwise you end up with `1.3%/year`

src/components/SubCategorySummary/SubCategorySummary.js
```diff
@@ -0,0 +1,74 @@
...
+import aggregateIndicatorForCities from '../../helpers/aggregateIndicatorForCities';
+import getSubCategorySectionId from '../../helpers/getSubCategorySectionId';
+import { INDICATORS } from '../../constants';
+import style from './SubCategorySummary.scss';
+import COLORS from '../../style/_colors.scss';
+
+const SubCategorySummary = (props) => {
+  const backgroundColor = COLORS[`${props.colorName.toUpperCase()}_${props.colorVariation}`];
+
+  const firstThreeIndicators = props.indicatorIds.slice(0, 3);
```
micapam, line 13: So the indicator cards are dictated by the order in indicators.js - right? If so we should document that somewhere.

micapam (approved):
Two small change suggestions, not blockers

src/constants/indicators.js
davidgilbertson: Yep just dummy data.

src/components/SubCategorySummary/SubCategorySummary.js
davidgilbertson: This is still a mystery. We don't know how we're treating the 'key' indicator for the page (e.g. do we also show it in the sub-category?), and don't know how we're defining what will and won't show here.

PR #25, opened 2017-10-05
Fix missing/unused packages

davidgilbertson:
It's Tidy Fridy.

We had some packages that were referenced but weren't in `package.json`. I also bumped the Node version to what we were both using anyway.

Side note, you can now run npm packages without installing them locally with npx. So you can do cool things like `npx npm-check` and it'll run the handy `npm-check` package straight from the registry. I did this for fun times and realised we had some broken packages.


micapam (approved):
Not sure what the conventions are regarding dependencies; implication of the description seems to be that any packaged referred to in the codebase should also be explicitly mentioned in package.json (even if it's a dependency of another package) - right? 

Anyway, it runs fine, so LGTM.

davidgilbertson:
Yes, that's the convention, it was odd that there were packages used in the code that weren't in the `package.json`.

PR #26, opened 2017-10-06
Column charts instead of bar charts

micapam:
Ready for review. Caveats: no design for changed position of y axis description at narrow viewports. (We can circle back and revise this later.)

Replace AllCitiesCategory.js with [this gist](https://gist.github.com/micapam/91a480941ebc6c30d312c05353f2c7c0) to test out the charts.

Desktop, non-stacked:
image

Desktop, stacked:
image

Mobile, non-stacked:
image

Mobile, stacked: 
image


src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -46,22 +46,25 @@ function sortChartData(cities, indicator) {
...
+
   // The indicator data contains things like titles and descriptions. But these can
   // also be passed in explicitly (e.g. for stacked charts where there are more than one indicator)
   // so here we take the passed in value, or the value from the first indicator otherwise.
   const firstIndicator = INDICATORS[props.indicatorIds[0]];
 
   // check for indicators that don't have a numeric data type
   const hasNonNumericIndicators = props.indicatorIds.find(
-    indicator => INDICATORS[indicator].dataType !== DATA_TYPES.NUMBER,
+    indicatorId => INDICATORS[indicatorId].dataType !== DATA_TYPES.NUMBER,
```
davidgilbertson, line 60: Nice one.

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -46,22 +46,25 @@ function sortChartData(cities, indicator) {
...
 }
 
-const CityBarChart = (props) => {
-  // If more than one indicator is passed in, this becomes a stacked bar chart
+const CityColumnChart = (props) => {
+  // If more than one indicator is passed in, this becomes a stacked column chart
   const isStacked = props.indicatorIds.length > 1;
 
+  // To ensure readability for various config options
+  const narrowViewport = document.documentElement.clientWidth < 650;
```
davidgilbertson: Minor: `window.innerWidth` is shorter, more traditional, but generally getting the screen width to define the display in JavaScript is a bad idea, more comments below.

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -107,20 +120,22 @@ const CityBarChart = (props) => {
...
       type: 'category',
       categories: data.map(city => city.name),
+      labels: xAxisLabelsConfig,
     },
     yAxis: {
       ceiling: props.ceiling,
       labels: {
+        padding: 0,
+        x: 0,
+        y: 0,
```
davidgilbertson: `y: 3,` aligns better. But that's odd, because 3 should be the default. Or was zero on purpose so that the number sat above the line?

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -153,22 +168,29 @@ const CityBarChart = (props) => {
...
           <Legend
             // Legend is our own HTML so we can style and position it with CSS
             className={style.legendWrapper}
+            title={shortDescription}
             series={series}
           />
         </div>
       )}
+
+      {(!isStacked && narrowViewport) && (
```
davidgilbertson: If this should be conditional on screen size, the logic should be in CSS, since all the code here only runs when the page loads. E.g. it won't run when a user rotates a device.

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -73,26 +76,36 @@ const CityBarChart = (props) => {
...
 
   // we set a left margin on the chart explicitly
   // this means we can left align the title and the legend with the y axis
-  const marginLeft = 120;
+  const marginLeft = narrowViewport ? 15 : 120;
+
+  // We start losing visibility of x-axis labels at viewport widths lower than 550 
+  const xAxisLabelsConfig = narrowViewport
+    ? { rotation: -90, style: { fontSize: '9px' }, padding: 0 }
+    : { rotation: -45 };
```
davidgilbertson: This should use the [Highcharts responsive feature](http://api.highcharts.com/highcharts/responsive). The code above won't update when a user rotates/resizes their device.

davidgilbertson (approved):
Some comments, but not the end of the world if they're ignored.

PR #27, opened 2017-10-06
Feat/sc 74 update colors

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-74.


src/components/Card/Card.scss
```diff
@@ -21,69 +21,3 @@
...
-  }
-}
-
-.card__extra_color {
-  border-bottom-color: $color-extra--400;
-
-  &.card__category {
-    background: $color-extra--700;
-  }
-}
```
davidgilbertson, line 89: I ditched all of this because we know the colour value already in the component, so I just apply it directly as a style rather than this roundabout way.

The same approach won't work with the header category nav because you can't define hover states in styles.

src/helpers/getColorRange.js
```diff
@@ -0,0 +1,17 @@
+import color from 'color';
+
```
micapam, line 2: Nice method.

micapam (approved):
All looks fine to me.

PR #29, opened 2017-10-11
Feat/sc 20 compare

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-20

Should look pretty on all screen sizes and keyboard nav should look native.

The heading text does not match the designs because it doesn't exist in typography so will be sorted later.

You can use it like so in `<CityCategory>`:

```jsx
<CompareCities
  city={props.city}
  category={props.category}
  cities={props.cities}
  toggleCitySelected={props.toggleCitySelected}
/>
```

Some screenshots. Note that when we get short city names we might adjust the column width.

Desktop
image

localhost_8080_cairns_housing ipad

localhost_8080_cairns_housing ipad 1

Big iPhone 
localhost_8080_cairns_housing iphone 6 plus

Little iPhone
localhost_8080_cairns_housing iphone 5


src/components/CompareCities/CompareCities.scss
```diff
@@ -0,0 +1,61 @@
...
+    flex-direction: row;
+    align-items: center;
+    padding: 24px 30px;
+  }
+}
+
+.title {
+  @include type-indicator-card-title--large;
+
+  margin-bottom: 8px;
```
micapam, line 28: Title and subtitle look slightly misaligned to me. Subtitle a little too high on the page.

src/app.js
```diff
@@ -38,3 +38,15 @@ fetch(DATA_URL)
   .catch((err) => {
     console.error(err);
   });
+
+// we want to apply some special CSS only for users who are
+// using the keyboard to navigate (e.g. focus state for custom check boxes)
+// so we add a class to the body
+function handleFirstTab(e) {
+  if (e.keyCode === 9) { // the tab key
```
micapam, line 46: This is cool. But there's no facility to remove this state. After you tab once, you get the highlights etc. even when you use the mouse. Not a big deal but maybe mouse click could remove class & restore listener?

micapam (approved):
LGTM, couple of minor points

src/app.js
davidgilbertson: This is by design. Focus rings are harmless, so the extra complexity to remove the focus ring on mouse move, and then remove the mouse listener and reattach the keydown listener, then remove it again on tab and add the mouse listener again and so on a so forth is, I think, too much complexity for too little benefit.

I wrote a [whole blog post about it](https://hackernoon.com/removing-that-ugly-focus-ring-and-keeping-it-too-6c8727fefcd2) :)

src/components/CompareCities/CompareCities.scss
davidgilbertson: Fixed (`align-items: baseline;`)

PR #30, opened 2017-10-12
Feat/sc 18 city cat overview

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-18

Desktop 1920x1080
image

localhost-8080-greater-brisbane-liveability ipad 1

localhost-8080-greater-brisbane-liveability ipad

localhost-8080-greater-brisbane-liveability iphone 6 plus

localhost-8080-greater-brisbane-liveability iphone 6 1


src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
```diff
@@ -23,30 +23,32 @@ class CategoryNav extends Component {
 
   render() {
     return (
```
davidgilbertson, line 25: All I did here was add a `<div className={style.wrapper}>` wrapper, but git got confused so it looks like I changed a whole lotta stuff.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.scss
```diff
@@ -1,30 +1,39 @@
...
+
+  @include for-desktop-up {
+    height: $top-nav-height-desktop;
+    padding-left: 0;
+  }
+}
+
+.categoryNav {
+  display: flex;
   white-space: nowrap;
```
davidgilbertson, line 17: The nav is now contained to a max width and aligned with the body of the page

micapam (approved):
Can't see anything amiss here, except maybe if we were to consider the city header shippable as is then we'd want to hide the iconWrapper altogether. (Easy enough to change later if that does turn out to be the case.)

PR #31, opened 2017-10-12
Update datavizkit to avoid warning re. shouldComponentUpdate()

micapam:
A class in Datavizkit was giving warnings - just needed to switch `PureComponent` to `Component` in datavizkit and then upgrade datavizkit version. 

Ref: https://github.com/govau/datavizkit/commit/49c9ede469d4b6223c655c59495d2aebbe82ab60


davidgilbertson (approved):
While you're at it, to stop datavizkit complaining about unmet dependencies , do you want to bump the `peerDependency` values in `package.json` to `"react": ">=15.5.4",` and `"react-dom": ">=15.5.4",`

PR #32, opened 2017-10-16
Feat/sc 72 city overview

davidgilbertson:
This was originally logged as two tasks, but they're handled by the same code, so done in one task.
Jira: https://govausites.atlassian.net/browse/SC-72 and https://govausites.atlassian.net/browse/SC-10

1920x1080
image

1366x768
image

iPhone 8
localhost_8080_all-cities_overview iphone 8

iPad
localhost_8080_all-cities_overview ipad

iPad
localhost_8080_all-cities_overview ipad 1


public/data/smartCitiesData.json
```diff
@@ -1,7 +1,9 @@
 [
   {
```
davidgilbertson, line 2: This is the generated file. It's changed because of changes in `convert.js` and `cities.js`

scripts/convert.js
```diff
@@ -21,11 +21,12 @@ const indicators = Object.entries(INDICATORS).map(([key, indicator]) => ({
...
-        name: row.Cities,
-        id: kebabCase(row.Cities),
-        indices: {},
-      };
+      const city = CITIES[row.Cities];
+
+      if (!city) throw new Error(`${row.Cities} is not a known smart city`);
+
+      city.name = row.Cities;
+      city.indices = {};
```
davidgilbertson, line 29: We now mix in some static city data in the same way we get the static indicator data.

src/components/CategoryBanner/CategoryBanner.test.js
```diff
@@ -1,47 +0,0 @@
-import React from 'react';
```
davidgilbertson, line 1: Note deleted, renamed to `PageBanner`

micapam:
You've mentioned two tasks @davidgilbertson but linked twice to the same task (SC-72) - what's the other one ?

davidgilbertson:
Oops. SC-10. Updated.

src/components/OverviewSection/OverviewSection.js
```diff
@@ -0,0 +1,67 @@
...
+  );
+};
+
+const cityPropShape = {
+  description: PropTypes.string.isRequired,
+  indices: PropTypes.object.isRequired,
+  shortName: PropTypes.string.isRequired,
+};
+
+OverviewSection.propTypes = {
```
micapam, line 62: I feel a bit squeamish about this (and how it ripples into other components too). Not sure if I have a better solution though.

micapam: This might be of interest: https://github.com/airbnb/prop-types
(Specifically: https://github.com/airbnb/prop-types/blob/master/src/mutuallyExclusiveProps.js )

scripts/convert.js
```diff
@@ -21,11 +21,12 @@ const indicators = Object.entries(INDICATORS).map(([key, indicator]) => ({
...
-      const city = {
-        name: row.Cities,
-        id: kebabCase(row.Cities),
-        indices: {},
-      };
+      const city = CITIES[row.Cities];
+
+      if (!city) throw new Error(`${row.Cities} is not a known smart city`);
+
+      city.name = row.Cities;
```
micapam, line 28: Are we sure the long names in the spreadsheet are the ones we want to surface on the website? If not, we could add in `name` or `longName` to the attributes in `cities.js`, and just use row.Cities as the key to look up (then throw away)

micapam (approved):
LGTM, minor comments

scripts/convert.js
davidgilbertson: Good point, don't know. Every time I ask I get a different answer so maybe we revisit this later.

src/components/OverviewSection/OverviewSection.js
davidgilbertson: Yeah, I hear ya. But I think in practice there's nothing that could go wrong and make it into production without a dev-time warning, (if neither are passed in it fails loudly) so I don't think it's worth adding more complexity in the form of another package when it wouldn't have any material impact.

micapam: I was thinking more of the case where _both_ are passed.

davidgilbertson: Ah, gotcha. Yes then all sorts of unknown things would go wrong.

PR #33, opened 2017-10-16
Feat/sc 81 perf indicator ui

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-81

This is the most complex P in the history of the world. The rules as to its appearance are in the task.

Alignment is shoddy but will be replaced with an SVG icon in SC-60, the icon task.

localhost_8080_all-cities_overview ipad 2
localhost_8080_all-cities_overview ipad 3


src/components/CategoryIndicator/CategoryIndicator.js
```diff
@@ -38,6 +38,7 @@ const CategoryIndicator = (props) => {
           className={style.card}
           indicator={props.category.heroIndicatorId}
           color={categoryColor}
+          colorName={props.category.colorName}
```
micapam, line 41: Why are both color and colorName required ?

src/components/IndicatorTypeMark/IndicatorTypeMark.js
```diff
@@ -0,0 +1,57 @@
...
+  LARGE: 'LARGE',
+};
+
+// A "P" in a circle
+const IndicatorTypeMark = (props) => {
+  const performanceMarkColor = getColorVariant(props.colorName, '900');
+
+  let pColor;
+  if (props.isCategoryPage) {
+    pColor = COLORS.WHITE;
```
micapam, line 20: I can't see where this gets used....?

micapam (approved):
lgtm

src/components/IndicatorTypeMark/IndicatorTypeMark.js
davidgilbertson: Wow, nicely caught. I started out just falling back to CSS for color, but got half-way to explicitly setting the colour here for clarity.

Fixed.

src/components/CategoryIndicator/CategoryIndicator.js
davidgilbertson: Yeah it's a bit of a mess. The `color` is an actual hex code. It needs to be passed in explicitly because it's different for every sub-category, so the parent component must calculate it.

But also, we need to just take the 900 variant of whatever the current category color is (`colorName`) for the `<IndicatorTypeMark>`, and we don't want to calculate that multiple times and pass it down as a HEX code.

PR #34, opened 2017-10-16
Feat/sc 16 all cities cat charts

micapam:
Jira: <https://govausites.atlassian.net/browse/SC-16>

Preview: <http://smart-cities-sc-16.apps.staging.digital.gov.au>

Data is still pretty arbitrary, looks at 'jobs & skills' to see a few charts laid out (in the first subcat). Data will be tackled separately in another task. Icons are also missing.


src/components/AllCitiesCategoryContainer/AllCitiesCategory/AllCitiesCategory.js
```diff
@@ -1,40 +1,59 @@
...
+      {props.category.subCategories.map(subCategory => (
+        <SubCategorySummary
+          key={subCategory.name.replace(' ', '-')}
+          {...subCategory}
+          colorName={props.category.colorName}
+          cities={props.cities}
+        />
+      ))}
+
+      {isHeroInSubCategory || <div className={style.heroWrapper}>
```
micapam: This hero chart might be redundant, not 100% sure if hero charts will always be in a subcategory, yet. Once data task is done we might want to get rid of this bit.

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -70,7 +70,7 @@ const CityColumnChart = (props) => {
   // so this works for a normal or a stacked chart
   const series = props.indicatorIds.map((indicatorId, i) => ({
     index: props.indicatorIds.length - i, // reverse sort the series (to counteract Highcharts)
-    name: INDICATORS[indicatorId].name,
+    name: INDICATORS[indicatorId].shortDescription,
```
micapam, line 73: Not 100% sure on this either (whether to use short description or name) - will need to review once we have the data sorted and can see what needs to be done.

src/constants/categories.js
```diff
@@ -11,21 +11,57 @@ export default [
         name: 'Employment',
         tint: '040',
         shade: '600',
-        indicatorIds: [
-          'participationRate',
-          'gdp',
+        summaryIndicatorIds: [
```
micapam: Changed the structure of this file so that summary indicators are decoupled from charts. Also, charts can render with one or more indicators.

src/constants/indicators.js
```diff
@@ -111,24 +109,6 @@ export default {
     format: '0[.]0%',
     source: 'Dwelling type ratio, Other, Percentage of Dwellings',
   },
-  jobsWalk: {
```
micapam: Some of the indicators previously added are no longer present in the 'master' data file.

src/style/_colors.scss
```diff
@@ -165,6 +165,8 @@ $color-primary--900: #003a45;
   PLANNING_700: $color-planning--700;
   PLANNING_800: $color-planning--800;
   PLANNING_900: $color-planning--900;
+
+  GRID: $color-grid;
```
micapam: This variable was missing, which is why the charts had no horizontal grid lines showing.

src/style/_icons.scss
```diff
@@ -0,0 +1,8 @@
+@mixin iconWrapper {
```
micapam, line 1: Consolidated this stuff with a view to implementing properly when we do icon system (and to DRY up the placeholder code)

src/components/AllCitiesCategoryContainer/AllCitiesCategory/AllCitiesCategory.js
davidgilbertson: Minor: for consistency/readability you can start the DOM on the next line:
```jsx
{isHeroInSubCategory || (
  <div className={style.heroWrapper}>
    <CityColumnChart
      cities={props.cities}
      colorBase={props.category.colorName}
      colorVariation="500"
      indicatorIds={[props.category.heroIndicatorId]}
    />
  </div>
)}
```

src/components/AllCitiesCategoryContainer/AllCitiesCategory/AllCitiesCategory.js
```diff
@@ -1,41 +1,60 @@
...
-      colorName={props.category.colorName}
-      description={props.category.description}
-      indicator={props.category.heroIndicatorId}
-      title={props.category.name}
-      cities={props.cities}
-      isCategoryPage
-    />
+const AllCitiesCategory = (props) => {
+  const isHeroInSubCategory = props.category.subCategories.find(sc => (
+    flatten(sc.charts.map(chart => chart.indicatorIds)).includes(props.category.heroIndicatorId)
```
davidgilbertson: Minor: this is a bit difficult to decipher. It might be more readable broken up.

Also, to 'cast' to boolean you would normally do `!!`, unless you want to explicitly discount `null`, `false`, etc (which you don't in this case, since it can only return an object or undefined). Also minor, I prefer full words in variable names (rather than `sc`), but that's not a rule anywhere, so...

E.g.
```js
  const isHeroInSubCategory = !!props.category.subCategories.find((subCategory) => {
    const indicatorsFromAllCharts = flatten(subCategory.charts.map(chart => chart.indicatorIds));

    return indicatorsFromAllCharts.includes(props.category.heroIndicatorId);
  });
```

Also, pretty minor, you could do this with native JS `.reduce()` instead of `flatten` removing the need to look up what `flatten` does in order to understand what the code does.

src/components/SubCategoryDetails/SubCategoryDetails.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import getSubCategorySectionId from '../../helpers/getSubCategorySectionId';
+import CityColumnChart from '../CityColumnChart/CityColumnChart';
+import { INDICATORS } from '../../constants';
+import style from './SubCategoryDetails.scss';
+
+const SubCategoryDetails = (props) => {
```
davidgilbertson, line 8: Minor: if this just shows charts, `<SubCategoryCharts>`?

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -135,18 +140,26 @@ const CityColumnChart = (props) => {
...
+            },
+          },
           xAxis: {
             labels: {
               rotation: -90,
               padding: 0,
-              style: { fontSize: '9px' },
+              style: {
+                fontSize: '9px',
+              },
```
davidgilbertson, line 162: Much better :)

src/components/SubCategoryDetails/SubCategoryDetails.js
```diff
@@ -0,0 +1,76 @@
...
+  const sc = props.subCategory;
```
davidgilbertson: We should probably standardise on whether to initialise variable names or not. It isn't done elsewhere in smart-cities, but is done in ICT. My preference is to keep the same name everywhere, so if the prop name is `subCategory` in one place, it is always `subCategory`.

Thoughts?

src/components/SubCategoryDetails/SubCategoryDetails.js
```diff
@@ -0,0 +1,76 @@
...
+
+  const heroChart = sc.charts.find(chart => (
+    chart.indicatorIds.includes(props.heroIndicatorId)));
+
+  const otherCharts = sc.charts.filter(chart => (
+    !chart.indicatorIds.includes(props.heroIndicatorId)));
```
davidgilbertson: Minor, prefer closing brace on new line:
```js
  const heroChart = sc.charts.find(chart => (
    chart.indicatorIds.includes(props.heroIndicatorId)),
  );

  const otherCharts = sc.charts.filter(chart => (
    !chart.indicatorIds.includes(props.heroIndicatorId)),
  );
```

src/components/SubCategoryDetails/SubCategoryDetails.js
```diff
@@ -0,0 +1,76 @@
...
+
+  return (
+    <div
+      key={sc.name}
+      id={getSubCategorySectionId(sc.name)}
+      className={style.container}
+    >
+      <div className={style.heading}>
+        <span className={style.iconWrapper} />
+        <h3 className={style.title}>{sc.name}</h3>
```
davidgilbertson: If this is `<h3>` chart title should be `<h4>` now.

src/components/SubCategoryDetails/SubCategoryDetails.js
```diff
@@ -0,0 +1,76 @@
...
+      </div>
+      <div className={style.chartGrid}>
+        {heroChart && <div className={style.heroChartWrapper} key="hero">
```
davidgilbertson: Minor, DOM starting on new line to keep indenting looking good.

src/components/SubCategoryDetails/SubCategoryDetails.js
davidgilbertson: `.title` hasn't been defined anywhere, this doesn't match the designs

src/components/SubCategoryDetails/SubCategoryDetails.js
```diff
@@ -0,0 +1,76 @@
...
+        ))}
+      </div>
+    </div>
+  );
+};
+
+SubCategoryDetails.propTypes = {
+  subCategory: PropTypes.shape({
+    name: PropTypes.string.isRequired,
+    tint: PropTypes.string.isRequired,
```
davidgilbertson: Is tint used anywhere?

src/components/AllCitiesCategoryContainer/AllCitiesCategory/AllCitiesCategory.js
```diff
@@ -1,41 +1,60 @@
...
+      ))}
+
+      {isHeroInSubCategory || <div className={style.heroWrapper}>
+        <CityColumnChart
+          cities={props.cities}
+          colorBase={props.category.colorName}
+          colorVariation="500"
+          indicatorIds={[props.category.heroIndicatorId]}
+        />
+      </div>}
```
davidgilbertson: The font size and spacing look wrong (between the 'Infrastructure' row and the chart title), they don't match the designs:
image

We'd need a different title style for hero charts, right? And different spacing at the top?

src/components/SubCategoryDetails/SubCategoryDetails.scss
```diff
@@ -0,0 +1,45 @@
...
+  flex-basis: 49%;
+
+  @include for-phone-only {
+    flex-basis: 100%;
+  }
+}
+
+.heroChartWrapper {
+  flex-basis: 100%;
+}
```
davidgilbertson: Unless I'm missing something, there's no need for `flex-basis` here, all you want to do is set the width, right? (`flex-basis` defaults to `auto` which just resolves to the `width`. Interestingly, the spec actually recommends against using flex-basis directly.)

Also you can do 50% not 49%.

Also, if you're setting, then overriding a value, prefer 'mobile first'.

You can do:
```scss
.chartWrapper {
  width: 100%;

  @include for-tablet-portrait-up {
    width: 50%;
  }
}

.heroChartWrapper {
  width: 100%;
}
```

Also, the axis labels are cut off (very neatly). See 0%, 5%, 0%, 5% in the screenshot.

src/components/SubCategoryDetails/SubCategoryDetails.scss
```diff
@@ -0,0 +1,45 @@
+@import '~style/variables';
+
+.container {
+  display: flex;
+  flex-direction: column;
```
davidgilbertson: You don't need flex here, the children are all just block level elements going down the page.
Setting these three have no effect, so can be removed:
```css
display: flex;
flex-direction: column;
width: 100%;
```

src/components/SubCategoryDetails/SubCategoryDetails.scss
```diff
@@ -0,0 +1,45 @@
...
+
+  @include for-tablet-landscape-up {
+    padding-left: 32px;
+  }
+}
+
+.heading {
+  display: flex;
+  align-items: center;
+  width: 100%;
```
davidgilbertson: Is there a reason for `width: 100%`? Anything with a display of `block` or `flex` will be a 'block-level' element which will take all available width.

Unless there's a browser quirk I'm missing.

src/components/SubCategoryDetails/SubCategoryDetails.scss
```diff
@@ -0,0 +1,45 @@
...
+  margin-bottom: 16px;
+}
+
+.chartGrid {
+  width: 100%;
```
davidgilbertson: `width: 100%` can go

src/components/SubCategoryDetails/SubCategoryDetails.scss
davidgilbertson: The layout here looks a bit funky when one chart has a legend. Also the title is indented in quite a way and looks strange. Also I wanted to do something with big red arrows.
image

src/components/SubCategorySummary/SubCategorySummary.scss
```diff
@@ -28,12 +28,7 @@
 }
 
 .iconWrapper {
-  // TODO (davidg): implement real icons
-  border: 1px dashed $color-grey--500;
-  width: 80px;
-  height: 80px;
-  flex: none;
-  margin-right: 8px;
+  @include iconWrapper;
```
davidgilbertson, line 31: yeah, nice.

src/constants/categories.js
```diff
@@ -11,21 +11,57 @@ export default [
...
+              'unemploymentRateIndigenous',
+            ],
+          },
+          {
+            name: 'Participation rate',
+            indicatorIds: [
+              'participationRateGeneral',
+              'participationRateFemale',
+              'participationRateMale',
+            ],
```
davidgilbertson, line 48: This needs a mechanism to feed through the text that goes in the tooltip (`longDescription`). If you do this here:
```js
{
  name: 'Participation rate',
  longDescription: 'This is the text that will show in the tooltip for this chart',
  indicatorIds: [
    'participationRateGeneral',
    'participationRateFemale',
    'participationRateMale',
  ],
},
```

And then in `<SubCategoryDetails>` and wherever else you render the chart you do this:
```jsx
<CityColumnChart
  title={chart.name}
  longDescription={chart.longDescription}
  cities={props.cities}
  colorBase={props.colorName}
  colorVariation={sc.shade}
  indicatorIds={chart.indicatorIds}
/>
```

src/components/SubCategoryDetails/SubCategoryDetails.js
```diff
@@ -0,0 +1,76 @@
...
+
+  const heroChart = sc.charts.find(chart => (
+    chart.indicatorIds.includes(props.heroIndicatorId)));
+
+  const otherCharts = sc.charts.filter(chart => (
+    !chart.indicatorIds.includes(props.heroIndicatorId)));
+
+  return (
+    <div
+      key={sc.name}
```
davidgilbertson: `key` can go. It's only needed inside a loop (where it's in the actual context of the array, you don't need it if a component was originally rendered as the output of a loop, if that makes sense)

src/components/SubCategoryDetails/SubCategoryDetails.scss
```diff
@@ -0,0 +1,45 @@
+@import '~style/variables';
```
davidgilbertson, line 1: I have a bunch of comments below, all adding up to this as the final CSS (assuming a new `<div className={style.wrapper}>` at the top of `<SubCategoryDetails>`).
```scss
.wrapper {
  background: $color-white;
  padding-top: 32px;
  margin-bottom: 16px;
}

.container {
  @include container;
}

.heading {
  display: flex;
  align-items: center;
  margin-bottom: 16px;
}

.chartGrid {
  display: flex;
  flex-flow: row wrap;
}

.chartWrapper {
  width: 100%;

  @include for-tablet-portrait-up {
    width: 50%;
  }
}

.heroChartWrapper {
  width: 100%;
}

.iconWrapper {
  @include iconWrapper;
}
```

src/components/SubCategoryDetails/SubCategoryDetails.scss
```diff
@@ -0,0 +1,45 @@
...
+
+.container {
```
davidgilbertson: The charts fill the full width of the page. I believe everything should be contained to be max 1140px. I could be wrong though.

davidgilbertson: You'd need to do this in here and for the hero chart.

src/components/SubCategoryDetails/SubCategoryDetails.js
davidgilbertson: And `key="hero"` can go, key is only needed directly inside a loop.

davidgilbertson (changes requested):
Back to you

src/components/AllCitiesCategoryContainer/AllCitiesCategory/AllCitiesCategory.js
micapam: Oh yep, that is nicer.

src/components/SubCategoryDetails/SubCategoryDetails.js
micapam: I think this is hangover from original copy-paste from sub cat summary

src/components/AllCitiesCategoryContainer/AllCitiesCategory/AllCitiesCategory.js
micapam: I needed to tweak the styling, but the title style is no different for hero charts as far as I can see: https://projects.invisionapp.com/d/main#/console/11859769/257007528/inspect

PR #36, opened 2017-10-18
Feat/sc 78 context page

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-78

This adds 'context' as a new category. The context page is special in a few ways. Each 'sub category' background colour isn't just a different tint of the base colour, it's a specific shade from another category. So I've refactored the code to use explicit colours rather than define only the shade/tint.

I could rip the whole `colorName` thing out, since it always matches the ID, but this would probably cause nasty conflicts with SC-16 so I'll do it later.

Desktop:
image


src/components/SubCategorySummary/SubCategorySummary.scss
```diff
@@ -1,5 +1,9 @@
 @import '~style/variables';
 
+.wrapper {
+  border-top: 4px solid $color-white;
```
davidgilbertson, line 4: No strictly related, but added the little white gap that was added to the designs at some point.

src/components/PageBanner/PageBanner.js
```diff
@@ -38,16 +38,18 @@ const PageBanner = (props) => {
...
         </div>
 
-        <div className={style.indicatorTypeMarkWrapper}>
-          <IndicatorTypeMark
-            className={style.indicatorTypeMark}
-            colorName={props.colorName}
-            isCategoryPage={props.isCategoryPage}
-            size="LARGE"
-          />
+        {!props.isContextPage && (
```
micapam, line 41: Ah is this the usual syntax? I had assumed `boolean || <markup />`

micapam (changes requested):
Not sure what's causing this, but I notice that, in desktop view (n.b. on my small laptop, its screen seems to be - or emulate, since it's retina - 1275px in width), when I'm looking at a category, I can't see the 'Overview' button any more in the category nav - it gets tucked away behind the cities nav. It may well be this is not a new bug per se but just something that has been exposed by the change in elements on the screen:

image

Also, in mobile view, if I'm on a category midway through the list, so that the category nav scrolls to position, I can't see the little edge of the button off to the left any more, to show me that there are other categories offscreen ...

missing_button

src/components/PageBanner/PageBanner.js
davidgilbertson: I've only ever seen the && syntax used, so it reads more like an `if`. When I saw your `||` I almost commented but figured it doesn't really make any difference.

davidgilbertson:
Ah, the nav items were too wide for the screen at just over 1200px. So when you clicked a category, they would get scrolled to the right a bit and the left one hidden. I've fixed that with slightly smaller spacing for desktop, then large again for 'big desktop' (1440px+)

There was a second problem of padding on the left which I introduced last week some time. I've fixed that too.

The width at 1200px now:

image

micapam (approved):

micapam:
Mobile & desktop views both look good now. lgtm

PR #39, opened 2017-10-27
Feat/sc 63 data

micapam:
Preview: <http://smart-cities-sc-63.apps.staging.digital.gov.au>


src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -97,7 +98,7 @@ const CityColumnChart = (props) => {
         pointWidth: 6,
       },
       series: {
-        stacking: 'normal',
+        stacking: props.stacked,
```
davidgilbertson: `series.stacking` as a boolean seems to work but is undocumented https://api.highcharts.com/highcharts/plotOptions.series.stacking so might stop working with an update.

You should probably do `stacking: props.stacked ? 'normal' : null,` instead.

src/components/Icon/iconDefinitions.js
```diff
@@ -304,6 +304,35 @@ export default {
...
+        <polygon points="2.46875 2.76036047 13.1669659 2.76036047 13.1669659 0.521988372 0.224431818 0.521988372 0.224431818 13.1320814 2.46875 13.1320814" />
+        <polygon points="73.3901023 13.1320814 75.6344205 13.1320814 75.6344205 0.521988372 62.6918864 0.521988372 62.6918864 2.76036047 73.3901023 2.76036047" />
+        <polygon points="2.46875 64.3164884 0.224431818 64.3164884 0.224431818 76.9247907 13.1669659 76.9247907 13.1669659 74.6864186 2.46875 74.6864186" />
+        <polygon points="73.3901023 74.6864186 62.6918864 74.6864186 62.6918864 76.9247907 75.6344205 76.9247907 75.6344205 64.3164884 73.3901023 64.3164884" />
+      </g>
+    ),
+  },
+  liveabilityPublicSafety: {
+    title: 'Liveability - Public safety',
+    viewBox: '',
```
davidgilbertson: This needs a `viewBox`. It will be in the original svg file (an attribute on the `<svg>` element). Is the `0 0 79 77` from the one above a copy/paste from within this file? Or does it match what was in the icon file?

src/components/Legend/Legend.js
```diff
@@ -5,7 +5,7 @@ import style from './Legend.scss';
 const Legend = props => (
   <div className={props.className}>
     {props.title && <div className={style.title}>{props.title}</div>}
-    {props.series.map(item => (
+    {props.series.reverse().map(item => (
```
davidgilbertson: `.reverse()` mutates the original array and mutating props is a no-no (it changes the sort order in other components - potentially, depending on rendering order and who knows what else).

So you first need to clone the array then sort the new array. Check out `sortChartData()` in `CityColumnChart.js` for an example (`.sort()` also mutates an array in place)

Actually, why reverse this anyway? Wouldn't we want the legend to be in the same order that the indicators are listed in the data? This is without sorting, seems the right way round to me:

image

But if you want it to sort differently on mobile, where it looks like this:
image


You should use `flex-direction: column-reverse` to do so at that screen size only.

src/constants/indicators.js
```diff
@@ -299,4 +329,424 @@ export default {
...
+    name: 'Languages other than English',
+    shortDescription: 'Language other than English spoken at home',
+    longDescription: 'Language other than English spoken at home, percentage of population',
+    dataType: DATA_TYPES.NUMBER,
+    aggregationMethod: AGGREGATION_METHODS.MEDIAN,
+    format: '0[.]0%',
+    contextual: false,
+    source: 'Language other than English spoken at home, All persons, % of population',
+  },
+};
```
davidgilbertson: I get a no-newline error when trying to run this.

src/constants/indicators.js
```diff
@@ -299,4 +329,424 @@ export default {
...
+  // perceivedSafety: {
+  //   name: 'Perceived safety',
+  //   shortDescription: 'TODO',
+  //   longDescription: 'TODO',
+  //   dataType: DATA_TYPES.NUMBER,
+  //   aggregationMethod: AGGREGATION_METHODS.MEDIAN,
+  //   format: '0[.]0%',
+  //   contextual: false,
+  //   source: 'TODO',
+  // },
```
davidgilbertson, line 449: Commented out code can go?

src/constants/indicators.js
```diff
@@ -299,4 +329,424 @@ export default {
...
+    contextual: false,
+    source: 'Building energy efficiency, Average rating',
+  },
+  volunteeringRate: {
+    name: 'Volunteering rate',
+    shortDescription: 'Percentage of population over 15 years who volunteer',
+    longDescription: 'Percentage of population over 15 years who volunteer',
+    dataType: DATA_TYPES.NUMBER,
+    aggregationMethod: AGGREGATION_METHODS.MEDIAN,
+    format: '0[.]0%',
```
davidgilbertson: Are all these formats correct, or do we have another task to do them?

src/constants/indicators.js
```diff
@@ -299,4 +329,424 @@ export default {
...
+    shortDescription: 'Particle pollution less than PM2.5',
+    longDescription: 'Air quality, annual mean: particle pollution less than PM2.5',
+    dataType: DATA_TYPES.NUMBER,
+    aggregationMethod: AGGREGATION_METHODS.MEDIAN,
+    format: '0[.]0a',
+    contextual: false,
+    source: 'Air quality, Annual mean, <PM2.5, ug/m3',
+  },
+  energyEfficiencyOfBuildings: {
+    name: 'Energy effiency of buildings',
```
davidgilbertson: Typo "efficiency"

src/constants/indicators.js
```diff
@@ -179,59 +179,89 @@ export default {
...
     contextual: false,
     source: 'Dwelling type ratio, Separate house, Percentage of Dwellings',
   },
   dwellingSemi: {
-    name: 'Semi-House',
-    shortDescription: 'TODO',
-    longDescription: 'TODO',
+    name: 'Dwelling type ratio: semi-detatched',
+    shortDescription: 'Semi-detatched',
+    longDescription: 'Dwelling type ratio: semi-detatched',
```
davidgilbertson: Three typos, "detached"

src/constants/indicators.js
```diff
@@ -179,59 +179,89 @@ export default {
...
+    dataType: DATA_TYPES.NUMBER,
+    aggregationMethod: AGGREGATION_METHODS.MEDIAN,
+    format: '0[.]0%',
+    contextual: false,
+    source: 'Education attainment - Completed Certificate III or IV, % of population',
+  },
+  educationalAttainmentTertiary: {
+    name: 'Educational attainment - Tertiary',
+    shortDescription: 'TODO',
+    longDescription: 'TODO',
```
davidgilbertson: Shouldn't all TODOs be gone in this task? Otherwise we need a new task to fix them.

src/constants/indicators.js
```diff
@@ -31,18 +31,18 @@ export default {
...
     aggregationMethod: AGGREGATION_METHODS.MEDIAN,
     format: '0[.]0%',
     contextual: true,
     source: 'Indigenous status, % of population',
   },
   population10yr: {
-    name: 'Population 10 year growth rate',
-    shortDescription: 'Growth rate',
-    longDescription: 'TODO',
+    name: '10-year population growth rate',
```
davidgilbertson: We should raise that this is ambiguous. It should highlight that it's the average annual growth, not the growth over 10 years.

src/constants/indicators.js
```diff
@@ -179,59 +179,89 @@ export default {
     contextual: false,
     source: 'Education attainment - Completed Year 12, % of population',
   },
+  educationalAttainmentCertificate: {
+    name: 'Educational attainment - Certificate',
```
davidgilbertson: Should be sentence case, little c

src/constants/indicators.js
```diff
@@ -179,59 +179,89 @@ export default {
...
+    shortDescription: 'TODO',
+    longDescription: 'TODO',
+    dataType: DATA_TYPES.NUMBER,
+    aggregationMethod: AGGREGATION_METHODS.MEDIAN,
+    format: '0[.]0%',
+    contextual: false,
+    source: 'Education attainment - Completed Certificate III or IV, % of population',
+  },
+  educationalAttainmentTertiary: {
+    name: 'Educational attainment - Tertiary',
```
davidgilbertson: Little 't' for Tertiary

davidgilbertson: And maybe-but-probably-not worth adding to the readme "All text should be sentence case"

src/constants/indicators.js
```diff
@@ -179,59 +179,89 @@ export default {
...
+    format: '0[.]0%',
+    contextual: false,
+    source: 'Education attainment - Completed Bachelor Degree or Higher, % of population',
+  },
   dispersion: {
     name: 'Local government dispersion',
-    shortDescription: 'Local government dispersion',
-    longDescription: 'Local government dispersion',
+    shortDescription: 'LGAs per 100,000 persons',
+    longDescription: 'Number of Local Government Areas per 100 000 persons',
```
davidgilbertson: Should "Local Government Areas" be lower case?

src/constants/indicators.js
```diff
@@ -179,59 +179,89 @@ export default {
...
     aggregationMethod: AGGREGATION_METHODS.MEDIAN,
     format: '0[.]0%',
     contextual: false,
     source: 'Government Dispersion, Number of Local Government Areas per 100 000 persons',
   },
   dwellingHouse: {
-    name: 'House',
-    shortDescription: 'TODO',
-    longDescription: 'TODO',
+    name: 'Dwelling type ratio: house',
```
davidgilbertson: There's colons here, but hyphens doing the same job up in "Educational attainment - Tertiary". Should they be the same?

src/constants/indicators.js
```diff
@@ -179,59 +179,89 @@ export default {
...
     format: '0[.]0%',
     contextual: false,
     source: 'Dwelling type ratio, Semi-detached, Percentage of Dwellings',
   },
   dwellingApartment: {
-    name: 'Apartment',
-    shortDescription: 'TODO',
-    longDescription: 'TODO',
+    name: 'Dwelling type ratio: flat or apartment',
+    shortDescription: 'Flat / apartment',
```
davidgilbertson: Slashes shouldn't (?) have spaces around them.

src/constants/indicators.js
```diff
@@ -179,59 +179,89 @@ export default {
...
     aggregationMethod: AGGREGATION_METHODS.MEDIAN,
     format: '0[.]0%',
     contextual: false,
     source: 'Dwelling type ratio, Other, Percentage of Dwellings',
   },
+  dwellingPriceToIncomeRatio: {
+    name: 'Dwelling price to household income ratio',
+    shortDescription: 'Dwelling price to household income ratio',
+    longDescription: 'Dwelling price to household income ratio',
+    dataType: DATA_TYPES.NUMBER,
```
davidgilbertson: If the indicators are all done now and `dataType` is always `DATA_TYPES.NUMBER` then this logic can be removed (now or in a new Jira task).

src/constants/indicators.js
```diff
@@ -10,19 +10,19 @@ import {
...
     contextual: true,
     source: 'Population, All persons, People',
   },
   growthRate: {
-    name: 'The growthRate',
-    shortDescription: 'TODO',
-    longDescription: 'TODO',
+    name: 'Annual population growth rate',
+    shortDescription: 'Annual growth',
+    longDescription: 'Population growth rate, percentage per annum',
```
davidgilbertson: "per annum" language used here but "per year" used further down. Do they mean the same thing, should they be consistent?

src/constants/categories.js
```diff
@@ -254,20 +631,86 @@ export const CATEGORIES = [
...
+        highlightColorDark: 'PLANNING_500',
+        iconId: 'planningLocalGovernment',
+        summaryIndicatorIds: [
+          'indigenousPopulation',
+          'growthRate',
+          'population10yr',
+        ],
+        charts: [
+          {
+            name: 'Total population',
```
davidgilbertson, line 645: Is this redundant? The chart does `const title = props.title || firstIndicator.name;` so for anything with one indicator there's no need to pass in a name.

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -77,7 +76,9 @@ const CityColumnChart = (props) => {
 
   // We only want to show the short description as the chart title
   // if the chart is not stacked
-  const yAxisTitle = isStacked ? {} : { text: shortDescription };
+  const yAxisTitle = isMultiple ? {} : { text: shortDescription };
+
+  const ceiling = props.ceiling || (props.stacked ? 1 : null);
```
davidgilbertson: Ceiling can probably go entirely. It's only purpose was to manually pass in 1 for stacked charts. As is, it's never actually passed in to the component.

src/components/CityColumnChart/CityColumnChart.js
davidgilbertson: Also grouped charts look funny. Is that going to be part of another task?

image

davidgilbertson (changes requested):

src/components/CityColumnChart/CityColumnChart.js
micapam: It is passed in, though - line 117

src/components/Icon/iconDefinitions.js
micapam: Good pickup - I must have omitted to paste this in from the SVG file.

src/constants/indicators.js
micapam: In this task is fine - but should that also include removing references to this elsewhere in the code ?

src/constants/indicators.js
micapam: I believe they're correct.

src/constants/indicators.js
micapam: This is temporarily commented out until James can get the data in. The data needs to be there but isn't currently available in the csv.

src/constants/indicators.js
micapam: Colons are better, I'll standardise on that.

src/constants/indicators.js
micapam: I believe it's an Official Term that Means Something Specific.

src/components/CityColumnChart/CityColumnChart.js
micapam: Do you mean the fact that both series are identical? If so, that's a problem with the data for that chart (new business entrants and exits, right?), as other charts work fine e.g.: 

image

src/components/Legend/Legend.js
micapam: Huh! I didn't know JS arrays' `reverse()` method did that.

On stacked, it's much of a muchness, but the reason to reverse the order becomes clear when you look at grouped (rather than stacked) charts.

src/components/CityColumnChart/CityColumnChart.js
davidgilbertson: I was unclear :) I mean a value for `props.ceiling` is never passed in to `<CityColumnChart>`.

src/constants/categories.js
micapam: Best to be explicit. Indicators can be used in more than one chart.

src/constants/indicators.js
micapam: Good point.

src/components/CityColumnChart/CityColumnChart.js
davidgilbertson: No I mean the way the bars overlap at some screen sizes looks odd.

src/constants/indicators.js
davidgilbertson: heh

src/constants/indicators.js
davidgilbertson: Yep. For memory I was checking it in the conversion file

src/constants/indicators.js
micapam: Or just link to the content guide ?

src/constants/indicators.js
micapam: Also, what about AGGREGATION_METHODS? Is that still needed?

jamesbroadbent202:
On the Context page all data looks right but the vertical legend on Median House Price and SEIA charts have the highest number in the legend appearing twice. $1m and 1k respectively

src/components/Legend/Legend.js
davidgilbertson: Yep, it's a mixed bag which methods mutate the original object in JavaScript. I had to google `reverse()` to check :)

FYI do you know about `legendIndex` in Highcharts? You can control the order that way, but it's probably more complex.

src/constants/categories.js
davidgilbertson: Oh, of course. 👍

src/constants/indicators.js
davidgilbertson: There's a content guide?

src/constants/indicators.js
davidgilbertson: It's gone in my branch that's ready for you to review.

jamesbroadbent202:
The dwelling price to income ratio chart vertical legend is showing 0% to 2500%. The data is a ratio with the highest being 21.1.

PR #40, opened 2017-10-30
Sc 96 dual numbers

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-96

This task wound up covering the 'page legend' that shows the indicator icons and the min/max icons, and also updates the indicator cards to show the new "C" icon.


src/components/Card/IndicatorCard/IndicatorCard.js
```diff
@@ -46,24 +103,31 @@ const IndicatorCard = (props) => {
     >
       <p className={style.header}>{indicator.name}</p>
 
-      {!indicator.contextual && (
+      {!props.isContextPage && (
```
micapam, line 106: Is this conditional necessary? I thought we wanted icons to show everywhere?

src/components/Card/IndicatorCard/IndicatorCard.js
```diff
@@ -12,23 +13,16 @@ import stripPrefixAndSuffix from '../../../helpers/stripPrefixAndSuffix';
 import getColorVariant from '../../../helpers/getColorVariant';
 import style from './IndicatorCard.scss';
 
-const classnames = require('classnames/bind').bind(style);
+// we don't create a new component here because the number styles are
+// tightly bound to the styles of the card they're in
```
micapam, line 17: Fair enough that a new component isn't required, but is there a way to enforce or at least document the object passed to the function? I see JS doesn't have named args, but could you use parameter destructuring...?

src/components/Card/IndicatorCard/IndicatorCard.js
```diff
@@ -38,6 +32,69 @@ const IndicatorCard = (props) => {
...
 
+  let numberType;
+  if (numberProps.showDualNumbers) {
+    if (numberProps.isSecondNumber) {
+      numberType = style.secondNumber;
+    } else {
+      numberType = style.firstNumber;
+    }
+  } else {
+    numberType = style.onlyNumber;
```
micapam, line 43: Feels to me like this class, `onlyNumber`, is redundant.

micapam (approved):
Few non-blocking comments, approved

src/components/Card/IndicatorCard/IndicatorCard.js
davidgilbertson: It technically is, but the CSS is so convoluted I felt it added a lot to the readability to set the font styles explicitly for 'only number' rather than set a general font size then override it for the small/large numbers.

src/components/Card/IndicatorCard/IndicatorCard.js
davidgilbertson: Yeah destructuring is our named args. Maybe we add documentation to this, but as is it's like all other functions that aren't documented. So maybe we do a JSDoc task at some point.

src/components/Card/IndicatorCard/IndicatorCard.js
davidgilbertson: We do, but the decision to strip out the logic of not showing it on the context page came after I'd finished. And TBH strikes me as a decision with 40% chance of mind changing.

PR #42, opened 2017-11-01
Feat/sc 19 city charts

micapam:
City charts, with compare cities.

Some styling fixes in there too, and performance/context icons for charts. But I have left out 'Last updated' because that not available - I've made a new Jira for that.

Jira: https://govausites.atlassian.net/browse/SC-19

Preview: http://smart-cities-sc-19.apps.staging.digital.gov.au/


src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -218,16 +268,21 @@ const CityColumnChart = (props) => {
         </div>
       )}
 
-      <AbstractWidget config={config} />
+      <div className={style.chartWrapper}>
+        <AbstractWidget ref={(c) => { this.chartWidget = c; }} config={config} />
```
davidgilbertson: What's `ref={(c) => { this.chartWidget = c; }}` do? It throws and error and can go, no?

src/components/CityColumnChart/CityColumnChart.scss
```diff
@@ -6,12 +6,13 @@
 
 .titleWrapper {
   display: flex;
+  flex-wrap: wrap;
```
davidgilbertson: Is this as it should be, with the (?) on the next row?

image

src/components/CityColumnChart/CityColumnChart.scss
```diff
@@ -6,12 +6,13 @@
...
   align-items: center;
   line-height: 1;
+}
 
-  @include for-tablet-landscape-up {
-    margin-bottom: 24px;
-  }
+.indicatorTypeMark {
+  flex-basis: 100%;
```
davidgilbertson: What does this do? You should almost never need to use `flex-basis` (same with `flex-grow` and `flex-shrink`)

src/components/CityColumnChart/CityColumnChart.scss
```diff
@@ -23,13 +24,22 @@
...
+  margin: 0 7px 7px 0;
 
   @include for-tablet-landscape-up {
     font-size: 24px;
   }
 }
 
+.metaWrapper {
+  padding-top: 4px;
+  font-size: 0.8em;
```
davidgilbertson: We don't use ems. Should just be 12px.

src/components/CityColumnChart/CityColumnChart.scss
```diff
@@ -23,13 +24,22 @@
...
-  margin-right: 7px;
+  margin: 0 7px 7px 0;
 
   @include for-tablet-landscape-up {
     font-size: 24px;
   }
 }
 
+.metaWrapper {
+  padding-top: 4px;
```
davidgilbertson: Where possible avoid setting top spacing like this. Now you have a 7px margin bottom on the title that sits above + a 4px padding top here for a total 11px gap between the two. It gets messy. Better to set margin-bottom: 11px on the title, unless there's something I'm missing.

src/components/CityColumnChart/CityColumnChart.scss
```diff
@@ -23,13 +24,22 @@
...
+  font-size: 0.8em;
+
+  span {
+    margin-left: 10px;
+  }
```
davidgilbertson: a) using tags as selectors is bad practice, but mostly b) you don't need this, just put a margin right on the thing to the left (`.indicatorTypeMark`). The also means we're sticking to defining spacing on the right and the bottom which keeps things sane.

Also the P isn't aligned, so you can fix both at once with this (negative top margin shifts the element up)
```scss
.indicatorTypeMark {
  margin: -3px 8px 0 0;
}
```

src/components/CityColumnChart/CityColumnChart.scss
```diff
@@ -50,6 +60,12 @@
   }
 }
 
+.chartWrapper {
+  @include for-tablet-landscape-up {
+    margin-top: 24px;
+  }
+}
+
 :global(.cityColumnChartTooltipHeader) {
```
davidgilbertson: Axes are all zeros. Should be fixed or added as a bug

image

src/components/SubCategoryDetails/SubCategoryDetails.scss
```diff
@@ -12,13 +12,13 @@
...
   align-items: center;
-  margin-bottom: 16px;
   border-top: 1px solid $color-grey--200;
-  padding-top: 16px;
+  padding-top: 24px;
 
   h3 {
     font-size: 44px;
     font-weight: 400;
+    margin: 0;
```
davidgilbertson: In `_globalStyles.scss` there is a margin reset for header tags, you should add this there instead of here, e.g.
```scss
p,
figure,
h1,
h2,
h3 {
  margin: 0;
  font-size: inherit;
  font-weight: inherit;
}
```

Also, selecting tags directly is not great.

src/components/CityColumnChart/CityColumnChart.js
davidgilbertson: Maybe something to fix in this task, or not, but the chart has a big fat padding of 16px that is added to the site padding of 16px.

as is:
image

Padding gone:

image

src/components/CityColumnChart/CityColumnChart.scss
```diff
@@ -50,6 +60,12 @@
...
   margin-left: 8px;
   margin-right: 8px;
```
davidgilbertson: Either in this task or as a bug, I think the section and chart title line-heights could do with some love:

image

On the chart it's too small and on the section it's too large.

package.json
```diff
@@ -20,7 +20,7 @@
   },
   "dependencies": {
     "@atlaskit/tooltip": "^6.0.0",
-    "@gov.au/datavizkit": "^1.0.34",
+    "@gov.au/datavizkit": "^1.0.38",
```
davidgilbertson: [shakes fist at datavizkit]

src/components/CityCategoryContainer/CityCategory/CityCategory.js
```diff
@@ -31,6 +33,25 @@ const CityCategory = props => (
...
+
+    <CompareCities
+      category={props.category}
+      city={props.city}
+      cities={props.cities}
+      toggleCitySelected={props.toggleCitySelected}
+    />
+
+    {props.category.subCategories.map(subCategory => (
+      <SubCategoryDetails
```
davidgilbertson: Maybe I mentioned elsewhere, these are really `<SubCategoryCharts>` now.

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -15,10 +16,16 @@ import {
 } from '../../constants';
 import style from './CityColumnChart.scss';
 
-function getSeriesDataForIndicator(cities, indicator) {
+function getSeriesDataForIndicator(cities, indicator, mainCity = null) {
```
davidgilbertson: ` = null` is redundant. `mainCity` will be `undefined` if not provided, which is just as falsey as `null`.

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -15,10 +16,16 @@ import {
...
   return cities.map((city) => {
     if (indicator in city.indices) {
-      return city.indices[indicator];
+      const val = city.indices[indicator];
+
+      if (mainCity && !city.selected && mainCity.id !== city.id) {
```
davidgilbertson: A comment of what's happening here might be good. It took me ten minutes to work it out :|

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -99,12 +124,21 @@ const CityColumnChart = (props) => {
...
       type: 'category',
       categories: data.map(city => city.name),
+      selectedCategory,
       labels: {
         rotation: -45,
         style: {
           fontSize: '10px',
         },
+        formatter() {
+          if (this.value === this.axis.userOptions.selectedCategory) {
```
davidgilbertson: If this just passes in `selectedCategory` only to read it back out as `this.axis.userOptions.selectedCategory` why not just skip the middle man? Especially since 'selectedCategory' isn't a real property which makes it a bit confusing (I went and lookup up the API to see what it did).

Also I think the style is off, and shouldn't be `em`. So, all together:

```js
formatter() {
  // on the city page, highlight the city name bold if it's the current city
  if (mainCity && this.value === mainCity.name) {
    return `<span style="font-weight: 700; font-size: 12px">${this.value}</span>`;
  }

  return this.value;
},
```

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -204,6 +242,18 @@ const CityColumnChart = (props) => {
...
       </div>
 
+      <div className={style.metaWrapper}>
+        <Icon
+          className={style.indicatorTypeMark}
+          color={colorDark}
+          icon={firstIndicator.contextual ? 'contextualIndicator' : 'performanceIndicator'}
+          size={14}
+        />
+        <span>
```
davidgilbertson: `<span>` can go if you're no longer targeting it (see comments elsewhere)

davidgilbertson (changes requested):

package.json
micapam: lol

src/components/CityCategoryContainer/CityCategory/CityCategory.js
micapam: They also have the header icon and text but I guess the main point of them is to house the charts, so yeah ...

src/components/CityColumnChart/CityColumnChart.js
micapam: Yeah, this is simpler, thanks

src/components/CityColumnChart/CityColumnChart.js
micapam: I'll remove the padding now. That ref is a hangover from when I thought I needed to have access to the chart instance, no longer required now.

src/components/CityColumnChart/CityColumnChart.scss
micapam: Those particular indicators had the wrong formats, I've fixed those now. However, some other ones still have trouble (e.g. 'Residential building approvals and growth'). I'm going to leave fixing those because I believe one of the changes incoming in design review is to shift out format into a header for the Y axis - which, though it would likely fix this particular issue, is something that needs negotiating & making sure it makes sense across the board.

src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -104,7 +131,16 @@ const CityColumnChart = (props) => {
         style: {
           fontSize: '10px',
         },
+        formatter() {
+          // on the city page, show the current city name in bold
+          if (mainCity && this.value === mainCity.name) {
+            return `<span style="font-weight: 700; font-size: 12px">${this.value}</span>`;
```
davidgilbertson, line 138: FYI, if you're wondering, I've added 12px/700 to the style guide in my branch (not committed yet) as discussed with Nick.

src/components/CityColumnChart/CityColumnChart.scss
```diff
@@ -23,13 +19,21 @@
 .title {
   font-size: 22px;
   font-weight: 300;
-  margin-right: 7px;
+  margin: 0 7px 11px 0;
```
davidgilbertson, line 22: This breaks the alignment of the (?) which got merged in yesterday:

image

You could wrap the title and the (?) in a div and give it the bottom margin. But I say let's get this merged and do that later.

davidgilbertson (approved):

PR #44, opened 2017-11-02
Feat/sc 103 non chart tooltip

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-103

Why not use a tooltip package from npm?
* It wouldn't save time.
* The Atlassian tooltip is not easy to style (the method used on ICT to hijack their class name doesn't work in smart cities because we've gone all it with CSS modules and so don't know the class name).
* the most popular "react tooltip" module was put out to pasture, just as most React UI packages are eventually.
* creating one from scratch only took a few hours, using the hot new https://reactjs.org/docs/portals.html


src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,94 @@
...
+    this.setState({ visible: false });
+  }
+
+  render() {
+    const { props, state } = this;
+
+    return (
+      <span // a span so it's valid HTML no matter where it's used
+        onMouseOver={this.showTooltip}
+        onMouseOut={this.hideTooltip}
```
micapam, line 65: This works great on desktop but on mobile you can end up with a floating, out-of-context tooltip:

image

It can be dismissed with a touch, but would be nice to just hide it on scroll, like this: 

```
  componentDidMount() {
    window.addEventListener('scroll', this.hideTooltip);
  }

  componentWillUnmount() {
    window.removeEventListener('scroll', this.hideTooltip);
  }
```

micapam (approved):
Minor UI improvement, otherwise good to merge

src/components/Tooltip/Tooltip.js
davidgilbertson: Holy crap why didn't I think of that?! I messed around for ages listening for touchstart to try and get it to hide and it was being a pain.

davidgilbertson: Done, FYI I add/remove it only when showing/hiding a tooltip (otherwise you get 20 scroll listeners when all the tooltips mount).

PR #45, opened 2017-11-05
Feat/sc 75 typography update

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-75

I've updated the readme, should be self explanatory (if not I'll update it more).

It's a bit of a weird one since none of the designs actually have the correct font styles, so there's nothing to check against.


src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
```diff
@@ -156,15 +151,17 @@ $minister-image-width: 214px;
...
 }
 
 .subInfoBodyText,
 .infoBodyText {
-  font-size: 22px;
+  font-size: 20px;
+  font-weight: 300;
   line-height: 37px;
-  letter-spacing: -0.7px;
+  letter-spacing: 1px;
```
micapam, line 164: The letters look way too widely spaced to me. 

image

micapam (changes requested):
One style needs fixing, otherwise good to merge

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
davidgilbertson: What browser is that? It doesn't look right.

Here's how it looks for me, Chrome on Mac and Edge and Chrome on Windows all looked fine. The designs are overlaid (the lighter text)

image

micapam: Chrome 61 on Mac.

davidgilbertson: Can you screenshot the whole screen? This is Chrome 61 macOS for me:
image

micapam (approved):

PR #47, opened 2017-11-06
Make component tests independent of data

davidgilbertson:
This change means we can change `indicator.js` and `categories.js` without affecting snapshots.


src/components/CityColumnChart/__snapshots__/CityColumnChart.test.js.snap
```diff
@@ -4,12 +4,12 @@ exports[`should match Snapshot 1`] = `
 "<div className=\\"wrapper\\">
   <div className=\\"titleWrapper\\">
     <h4 className=\\"title\\">
-      Population
+      The population indicator name
     </h4>
-    <AboutTooltip text=\\"Total population\\" color=\\"JOBS_600\\" className=\\"aboutChartIcon\\" />
+    <AboutTooltip text=\\"The population indicator long description\\" color=\\"JOBS_600\\" className=\\"aboutChartIcon\\" />
```
micapam: This is actually better anyway as these strings are self-documenting

micapam (approved):
LGTM

PR #53, opened 2017-11-07
Fix for error in convert script

micapam:
This PR contains the following: 
- Fix for the convert script (it was still trying to look up number formats which no longer existed) 
- A test for the convert script. It's not the prettiest, I had some asynchronously capturing stdout - I managed it, but suggestions on a more elegant solution welcome. The one thing I am happy with is that I didn't have to distort the convert script itself to fit the testing framework.
- Update to the data
- Change to the deploy script to ensure that if the data conversion fails, it won't deploy


scripts/convert.js
```diff
@@ -33,19 +31,13 @@ getStdin().then((rawCsv) => {
...
-          } else {
-            // otherwise just use the value as text
-            city.indices[indicator.key] = value;
+          // All data should be numeric
+          if (Number.isNaN(Number(value))) {
+            throw new Error(`value of "${value}" for ${row.Cities} for ${indicator.source} is not a number`);
           }
+
+          // note this will convert blank cells to 0
+          city.indices[indicator.key] = Number(value);
```
davidgilbertson, line 40: This is probably a good opportunity to handle empty cells.

scripts/convert.test.js
```diff
@@ -0,0 +1,70 @@
...
+    },
+    'Greater Perth': {
+      id: 'greater-perth',
+      shortName: 'Perth',
+      description: 'Description of Perth',
+    },
+  },
+}));
+
+const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms));
```
davidgilbertson: This can be shortened (and moved into the test if it's only used once). FYI all of these are the same:
```js
await new Promise(resolve => setTimeout(resolve, 1));
await new Promise(resolve => setTimeout(resolve, 0));
await new Promise(resolve => setTimeout(resolve));
await new Promise(setTimeout);

// and also...
await new Promise(process.nextTick);
```

About `nextTick()` if you're curious: https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/

scripts/convert.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+const mockActualOutput = [];
+const originalWrite = process.stdout.write;
+
+jest.mock('get-stdin', () => () => {
+  return new Promise(resolve => {
+    process.stdout.write = (str) => mockActualOutput.push(str);
+    resolve(mockCsvData);
+  });
+});
```
davidgilbertson: There's a shorthand for resolving promises like so:
```js
jest.mock('get-stdin', () => () => Promise.resolve(mockCsvData));
```

scripts/convert.test.js
```diff
@@ -0,0 +1,70 @@
...
+    "id": "greater-perth",
+    "shortName": "Perth",
+    "description": "Description of Perth",
+    "name": "Greater Perth",
+    "indices": {
+      "population": 2040000,
+      "growthRate": 0.016
+    }
+  }
+]`;
```
davidgilbertson: I'm 50/50 on this, but rather than testing that the output is an exact string, you could test that parsing the output results in a particular object. Downside: you're not directly testing the output. Upside: you're testing that the output is valid JSON and still testing that it output the correct data.

scripts/convert.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+const mockActualOutput = [];
+const originalWrite = process.stdout.write;
```
davidgilbertson: Rather than overwriting the behaviour of `write()` to push to an array, you can instead make it a mock and check what it was called with. So 
```js
process.stdout.write = jest.fn();
```
and then:

```js
expect(process.stdout.write).toHaveBeenCalledWith(mockExpectedOutput);
```

scripts/convert.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+jest.mock('get-stdin', () => () => {
+  return new Promise(resolve => {
+    process.stdout.write = (str) => mockActualOutput.push(str);
```
davidgilbertson: (FYI, this is actually what `jest.fn()` does in the background, and then makes the result available on the `.mock.calls` object, and it's what `toHaveBeenCalledWith()` accesses.)

scripts/convert.test.js
```diff
@@ -0,0 +1,70 @@
+const mockCsvData = `Cities,"Population, All persons, People","Population, persons, Annual Growth %"
```
davidgilbertson, line 1: Some comments below, all adding up to this:
```js
const mockCsvData = `Cities,"Population, All persons, People","Population, persons, Annual Growth %"
Greater Sydney,4920000,0.021
Greater Perth,2040000,0.016`;

const expectedOutputObject = [
  {
    id: 'greater-sydney',
    shortName: 'Sydney',
    description: 'Description of Sydney',
    name: 'Greater Sydney',
    indices: {
      population: 4920000,
      growthRate: 0.021
    }
  },
  {
    id: 'greater-perth',
    shortName: 'Perth',
    description: 'Description of Perth',
    name: 'Greater Perth',
    indices: {
      population: 2040000,
      growthRate: 0.016
    }
  }
];

jest.mock('../src/constants', () => ({
  INDICATORS: {
    population: {
      name: 'The population indicator name',
    },
    growthRate: {
      name: 'Annual population growth rate',
      source: 'Population, persons, Annual Growth %',
    },
  },
  CITIES: {
    'Greater Sydney': {
      id: 'greater-sydney',
      shortName: 'Sydney',
      description: 'Description of Sydney',
    },
    'Greater Perth': {
      id: 'greater-perth',
      shortName: 'Perth',
      description: 'Description of Perth',
    },
  },
}));

jest.mock('get-stdin', () => () => Promise.resolve(mockCsvData));

process.stdout.write = jest.fn();

it('Converts the data correctly', async () => {
  require('./convert');

  await new Promise(setTimeout);

  const rawOutput = process.stdout.write.mock.calls[0];

  expect(JSON.parse(rawOutput)).toEqual(expectedOutputObject);
});
```

scripts/convert.test.js
```diff
@@ -0,0 +1,70 @@
...
+      description: 'Description of Perth',
+    },
+  },
+}));
+
+const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms));
+
+it('Converts the data correctly', async () => {
+  require('./convert');
+  await sleep(1); // This horrific line is because I can't `await` running the script directly
```
davidgilbertson: For future reference... using `done()`  _should_ work, but doesn't because of a bug in Jest...
```js
it('Converts the data correctly', done => {
  require('./convert');

  setTimeout(() => {
    expect(process.stdout.write).toHaveBeenCalledWith(mockExpectedOutput);
    done();
  }, 100);
});
```

davidgilbertson (changes requested):
Tests could be tidied up, you can copy/paste or pick and choose...

scripts/convert.test.js
micapam: Ahhhh!

scripts/convert.test.js
micapam: This is much cleaner, thanks. I think the only thing missing is to clean up the mock of `process.stdout.write` ... or does that happen automatically at the end of this file's test?

scripts/convert.test.js
micapam: Ahhh that's great. I use the equivalent in `rspec-mocks`, I was searching in jest documentation but hadn't found this!

PR #55, opened 2017-11-08
Feat/new cloud env

micapam:
Some 🔧 to prepare for prod deploys. Doesn't really change much now, but once it's in we will also switch to the new cloud env for dev & staging


davidgilbertson (approved):
No idea, looks great!

PR #60, opened 2017-11-10
Feat/sc 112 prod deploys

micapam:
Resolves SC-112: https://govausites.atlassian.net/browse/SC-112

This switches from plain jobs config to the new Circle 2 'workflows' config. 

Both staging and prod are now on their respective new Cloud environments.

Deployment instructions in readme.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,4 +72,4 @@ workflows:
                 /.*/
             tags: 
               only: 
-                /^v\d/
+                /^v[0-9]/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,4 +71,4 @@ workflows:
               ignore: 
                 /.*/
             tags: 
-              only: /^v[0-9]/
+              only: /^v/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ workflows:
   version: 2
   build_test_deploy:
     jobs: 
-      - build
+      - build:
           filters:
             tags:
               only: /.*/ # i.e. always build for all tags & branches
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,4 +72,4 @@ workflows:
                 /.*/
             tags: 
               only: 
-                /^release-.*/
+                /^v\d/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,21 +44,21 @@ workflows:
     jobs: 
       - build
       - deploy_staging: 
-        requires: 
-          - build
-        filters: 
-          tags: 
-            ignore: 
-              /.*/
-          branches: 
-            only: /^master$/
+          requires: 
+            - build
+          filters: 
+            tags: 
+              ignore: 
+                /.*/
+            branches: 
+              only: /^master$/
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -2,43 +2,74 @@
...
-            fi
+  deploy_staging: 
+    docker: 
+      - image: circleci/node:8.1.2
+    working_directory: ~/repo
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+          - v1-dependencies-{{ checksum "package.json" }}
```
micapam, line 32: Can't find a way to DRY this up. It would be nice if Circle let jobs share steps.

.circleci/config.yml
```diff
@@ -2,43 +2,74 @@
...
+      - run: ./scripts/cf_production.sh
+      - run: cf zero-downtime-push smart-cities -f manifest.yml -p build
+workflows: 
+  version: 2
+  build_test_deploy:
+    jobs: 
+      - build:
+          filters:
+            tags:
+              only: /.*/ # i.e. always build for all tags & branches
```
micapam, line 59: This cryptic syntax is because the [rules for tag filtering](https://circleci.com/docs/2.0/workflows/#git-tag-job-execution) are different than for branch filtering. If there are no filters, Circle assumes you want to run the job on all branches, and no tags - so you have to have a wildcard 'filter' to force the project to build & test on tags (or it will never reach the deployment step).

README.md
```diff
@@ -115,5 +115,11 @@ only the `build` directory is deployed, so you should `npm build` before deployi
...
 to the `master` branch, it will (assuming tests pass) additionally deploy to the
-`smart-cities-dev` instance. To deploy to the `smart-cities-staging` instance,
-simply push to the `staging` branch.
+`smart-cities-staging` instance.
+
+Production deploys are triggered when a tag is pushed with a name in the form `release-*`.
+To deploy, simply: 
+
+```
+git tag `date "+release-%Y%m%d%H%M%S"` && git push --tags
```
davidgilbertson: Is it worth aligning how we do tagging with other web projects? Using `v1.0.0` format is common (e.g. [npm releases](https://github.com/npm/npm/releases)) and [GitHub recommends](https://help.github.com/articles/creating-releases/) using semver for tags.

There's an npm command for this: `npm version`. That bumps the version in `package.json` and creates a git tag too. https://docs.npmjs.com/cli/version

So the instructions would be: run `npm version major` and that bumps the version and commits it with the appropriate tag.

Thoughts?

davidgilbertson (approved):
Approved but maybe use `npm version` instead of `release-*`

README.md
micapam: Good idea. The format was copied from performance dashboard; since this is a purely JavaScript app, using npm would work. 

I notice that `npm version major` (or `minor`, etc) commits but doesn't push.

I'll also need to change the tag filter config to pick up that format. `/^v\d+\.\d+\.\d+/` should work, I think (or even just `/^v\d/`)

davidgilbertson: You could add an npm script `"deploy": "npm version major && git push --tags"` I think would work.

No point in doing anything other than major since it isn't code that has an API.

PR #61, opened 2017-11-13
Feat/sc 121 update indicator mapping

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-121

This implements the changes in this file: https://docs.google.com/spreadsheets/d/1viOGavDc5E8TH4wve9cA5FVljZcJDz2RbpumbtDC274/edit#gid=149996498
And also:
* Moves descriptions from `indicators.js` to `categories.js` since the descriptions relate to the chart, not the indicator.
* For the indicators that show on multiple-series charts, add legend text.
* Name the properties for text clearly (e.g. `axisTitle`, `legendText`)
* Make sure all charts that show on the context page are also shown on their own category page


src/constants/categories.test.js
```diff
@@ -1,9 +1,50 @@
...
 import iconDefinitions from '../components/Icon/iconDefinitions';
 
 const iconIds = Object.keys(iconDefinitions);
 
+const heroExistsInCharts = (heroIndicatorId, subCategories) => {
+  let result = false;
+
+  subCategories.forEach(subCategory => {
+    subCategory.charts.forEach(chart => {
+      if (chart.indicatorIds.includes(heroIndicatorId)) result = true;
```
micapam, line 15: `return true` here would save a few cycles

src/constants/categories.test.js
```diff
@@ -1,9 +1,50 @@
...
+
+export const indicatorExistsInCategory = (indicatorId) => {
+  let result = false;
+
+  CATEGORIES
+    .filter(category => category.id !== CATEGORY_IDS.CONTEXT)
+    .forEach(category => {
+      category.subCategories.forEach(subCategory => {
+        subCategory.charts.forEach(chart => {
+          if (chart.indicatorIds.includes(indicatorId)) result = true;
```
micapam, line 30: same here

micapam (approved):
Minor comments, LGTM

src/constants/categories.test.js
davidgilbertson: Returning here actually only returns for this iteration (it's returning the current anonymous function, not the parent parent parent function).

To break early out of loops (or return a function from within a loop) you need to use a `for` loop.

Anyhoo, this all takes about 0.009ms.

Oh FYI do you know about `console.time()` and `.timeEnd()`? It's an easy way to decide if optimising would be premature. Use like so:
```js
  console.time('loop and loop and loop');
  
  CATEGORIES
    .filter(category => category.id !== CATEGORY_IDS.CONTEXT)
    .forEach(category => {
      category.subCategories.forEach(subCategory => {
        subCategory.charts.forEach(chart => {
          if (chart.indicatorIds.includes(indicatorId)) result = true;
        });
      });
    });

  console.timeEnd('loop and loop and loop');
```

PR #63, opened 2017-11-13
Feat/114 additional info

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-114

Can be seen at http://localhost:8080/additional-information - it's not linked to yet, that'll be done in the footer task.


src/components/AdditionalInfoPage/AdditionalInfoPage.js
```diff
@@ -0,0 +1,127 @@
...
+        <h1 className={style.title}>Content credit</h1>
+
+        <p className={style.paragraph}>
+          City photo - Albury<br />
+          Tim J Keegan<br />
+          Link: <Link href={LINKS.ALBURY_PHOTO_CREDIT} external />
+        </p>
+
+        <p className={style.paragraph}>
+          Site Icons<br />
```
micapam: 'Icons' should be 'icons', I reckon.

src/components/Link/Link.test.js
```diff
@@ -0,0 +1,69 @@
...
+  it('should add rel noopener noreferrer', () => {
+    const component = shallow(
+      <Link  {...defaultProps} />
+    );
+
+    expect(component.prop('rel')).toBe(null)
+  });
+});
+
+describe('when the external flag is true', () => {
```
micapam: All use cases appear to be external. Why not make this behaviour the default (or at least, if the href is http(s))? Maybe with an 'internal' attribute to suppress it (though I can't see why we would want to use this component for an internal link).

micapam (approved):
couple of small suggestions, not blockers

src/components/AdditionalInfoPage/AdditionalInfoPage.js
davidgilbertson: Agreed. Fixed.

src/components/Link/Link.test.js
davidgilbertson: Good point. I have removed all unused complexity (it is always external now). Considered renaming it `<ExternalLink>` but did not.

micapam:
Just one small point on those last changes @davidgilbertson - the Link component doesn't check what kind of href it's been given, so the mailto: link for Contact Us ends up with extraneous attributes:

```
<a class="Link-link--33Bv" href="mailto:performance-dashboard@digital.gov.au"
rel="noopener noreferrer" target="_blank">contact us</a>
```

Still works, but it might be worth stripping them out for non-http(s) links.

davidgilbertson:
If you set Chrome to handle `mailto` links with gmail then the link behaves just like it does for `http` links, so `rel` and `target` do apply.

micapam:
Cool. Merge away!

PR #65, opened 2017-11-14
Feat/sc 80 footer

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-80

After only several weeks, the footer task is complete. Huzzah.


src/components/FooterContainer/Footer/Footer.js
```diff
@@ -0,0 +1,111 @@
...
+            </div>
+          </div>
+        )}
+
+        <div className={style.staticLinksWrapper}>
+          {links.map(link => (
+            <a
+              key={link.href}
+              href={link.href}
+              target="_blank"
```
micapam: Why do these links open in new tab/window?

src/components/FooterContainer/Footer/Footer.js
```diff
@@ -0,0 +1,111 @@
...
+
+        <img
+          className={style.dtaLogo}
+          src={dtaLogo}
+          alt="dta"
+        />
+      </div>
+
+      <div className={style.wordsWrapper}>
+        {!!props.category && (
```
micapam, line 62: This seems bizarre. Why are these category/subcategory links included here? :S

micapam (changes requested):

src/components/FooterContainer/Footer/Footer.js
micapam: (This is not a change suggestion just a comment)

davidgilbertson: I concur, and can't answer that. But it increased the complexity significantly so I was happy to do it.

src/components/FooterContainer/Footer/Footer.js
davidgilbertson: I thought it was defined somewhere, but I don't see it in the task. I'll remove it.

davidgilbertson: Fixed

micapam (approved):

PR #66, opened 2017-11-14
Feedback page

micapam:
Resolves SC-119.


src/components/FeedbackPage/FeedbackPage.scss
```diff
@@ -0,0 +1,31 @@
...
+}
+
+.title {
+  margin-bottom: 40px;
+  font-size: 44px;
+  font-weight: 300;
+}
+
+.paragraph {
+  font-size: 22px;
```
davidgilbertson, line 19: This should be `font-size: 300` (as one of the 'allowed styles' in`/style-guide` in our extremely elegant typography system)

src/components/FeedbackPage/FeedbackPage.scss
```diff
@@ -0,0 +1,31 @@
+@import '~style/variables';
+
+.feedbackWrapper {
+  @include container;
+
+  padding: 48px 0;
+  background: $color-overview--020;
+}
+
```
davidgilbertson: I think there should be a wrapper, with the background colour, and a nested container (with max-width). Otherwise the colour doesn't go to the edge of the page and looks weird.

image

src/components/FeedbackPage/FeedbackPage.scss
```diff
@@ -0,0 +1,31 @@
+@import '~style/variables';
+
+.feedbackWrapper {
+  @include container;
+
+  padding: 48px 0;
```
davidgilbertson: This overrides the left/right padding of the container so it gets smooshed up against the sides
image

But is fixed by having a separate container and wrapper, see below.

src/components/FeedbackPage/FeedbackPage.scss
```diff
@@ -0,0 +1,31 @@
...
+.surveyFrame {
+  margin-top: 60px;
+  border: 1px solid $color-grey--500;
+  border-radius: 4px;
+  min-width: 100%;
+  min-height: 500px;
+
+  @include for-tablet-portrait-up {
+    min-width: 600px;
+    max-width: 800px;
```
davidgilbertson: Is this as per the survey monkey specification? Could we simplify this, do away with the breakpoint and do:
```css
  width: 600px;
  max-width: 100%;
```

src/components/FeedbackPage/FeedbackPage.scss
```diff
@@ -0,0 +1,31 @@
+@import '~style/variables';
```
davidgilbertson, line 1: I got carried away again, here's the result of my comments below:
```scss
@import '~style/variables';

.feedbackWrapper {
  padding: 48px 0;
  background: $color-overview--020;
}

.container {
  @include container;
}

.title {
  margin-bottom: 40px;
  font-size: 44px;
  font-weight: 300;
}

.paragraph {
  font-size: 22px;
  font-weight: 300;
}

.surveyFrame {
  width: 600px;
  max-width: 100%;
  min-height: 500px;
  margin-top: 60px;
  border: 1px solid $color-grey--500;
  border-radius: 4px;
}
```

davidgilbertson (changes requested):
A few changes...

src/components/FeedbackPage/FeedbackPage.scss
micapam: Hmm... `font-weight`, right?

davidgilbertson (approved):
A minor thing, there's no point in doing `min-height` on an iframe, it won't stretch based on content. You could just change to `height`

PR #68, opened 2017-11-15
Feat/sc 21 design updates

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-21


scripts/convert.js
```diff
@@ -23,7 +23,6 @@ getStdin().then((rawCsv) => {
 
       if (!city) throw new Error(`${row.Cities} is not a known smart city`);
 
-      city.name = row.Cities;
```
davidgilbertson, line 26: We ignore the spreadsheet version and just use the city object name.

micapam:
maybe get rid of shortName from CityOverview.test.js as it's been removed everywhere else

micapam (approved):
one tiny comment, lgtm

davidgilbertson:
Nicely spotted. Fixed

PR #71, opened 2017-11-16
Fix colours

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-126

Before, there were 'primary', 'overview' and 'context' colours (50 shades of teal). These are all now a single range. Also the mechanism for getting a range of colours for a chart is new.


src/helpers/getColorRange.js
```diff
@@ -1,17 +1,22 @@
-import color from 'color';
+import getColorVariant from './getColorVariant';
+import { COLOR_NAMES } from '../constants';
 
-export default function getColorRange(startColorCode, numberOfColours) {
-  const darkenBy = (1 / numberOfColours) + 0.15; // magic number, just tweaked till it looked nice
+// This is only ever used by charts, hence the lack of a CONTEXT set of colors.
+const CATEGORY_COLOR_MAP = {
```
micapam, line 5: urgh

micapam (approved):
the horror. but lgtm

src/helpers/getColorRange.js
davidgilbertson: I know, it hurts.

PR #75, opened 2017-11-17
Split manifest for prod vs staging so that we can specify the domain for prod

aeijdenberg:
Enable tailing of app during deployment, this is useful for debugging any deploy errors.


aeijdenberg:
Note since this is a CI change - it's a bit hard to test locally, so we may have to just keep an eye on it in staging...

manifest-production.yml
```diff
@@ -0,0 +1,8 @@
+---
+inherit: manifest-base.yml
+
+domain: dashboard.gov.au
```
micapam: Shouldn't this domain be `smart-cities.dashboard.gov.au` ?

micapam (changes requested):

manifest-production.yml
aeijdenberg: One would think so, but no, it actually is the parent domain, where the app name ends up sitting as a prefix underneath.

Confusing right? :)

micapam: Ah I see, thanks. Maybe drop in a quick comment to that effect in the file?

aeijdenberg: Done

micapam (approved):

PR #77, opened 2017-11-17
Basic GA (track uniques)

micapam:


public/index.html
```diff
@@ -1,6 +1,14 @@
...
 <html lang="en">
   <head>
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-61222473-12"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'UA-61222473-12');
+    </script>
```
davidgilbertson: Minor: for best performance, scripts should never be in the head, they should always go at the bottom of the page (e.g. below `<div id="root"></div>`). Although in reality, for this project, it will make no difference.

There is also an [old vulnerability](http://michaelthelin.se/security/2014/06/08/web-security-cross-site-scripting-attacks-using-utf-7.html) for docs that don't have the `charset` declaration in the first 1024 bytes, so putting stuff at the top of the head is a bad habit for that reason too. Again, in reality, probably none of our users have browsers that old, and it's probably still within the first 1024 bytes anyway.

(FWIW, Google is bad, and wrong, for saying it needs to go at the top of the head.)

davidgilbertson (approved):
Approve but consider moving script down.

PR #78, opened 2017-11-17
City nav links should point to overview

micapam:
Jira: https://govausites.atlassian.net/browse/SC-133


src/components/CityNavContainer/CityNav/CityNav.js
```diff
@@ -11,12 +11,15 @@ const CityNav = (props) => {
     [style.citySelected]: props.cityId !== NO_CITY,
   });
 
+  const locationHasCity = (location, city) => location && location.pathname.startsWith(`/${city}/`);
```
davidgilbertson: Technically this would throw if `location.pathname` was undefined.

src/components/SideBar/SideBar.js
```diff
@@ -49,7 +49,6 @@ class SideBar extends Component {
         </NavLink>
 
         <CityNavContainer
-          categoryId={props.categoryId}
```
davidgilbertson: `categoryId` can go from propTypes in this file, and removed from `SideBar.test.js` and no longer needs to be passed down to `<SideBar>` from `<PageWrapper>`.

Also, the comment in the line below is a leftover and can go.

davidgilbertson (changes requested):
_This_ should have some vestigial code removed.

micapam:
back to you @davidgilbertson

davidgilbertson (approved):
Oops, this dropped off my radar

PR #82, opened 2017-11-20
SC-134 data bugs part 2

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-134

There's not a spec for the number formatting, I've just done what works so most numbers are three figures. e.g.
* two decimal places for indicators that are mostly less than 10 (Air pollution greater than PM10)
* one decimal place for mostly-double-digit indicators
* and so on


src/constants/indicators.js
```diff
@@ -221,37 +221,37 @@ export default {
...
+    format: '0[.]0[0]%',
     contextual: false,
     lastUpdated: '2017',
     source: 'Unemployment rate, Persons, % of population by sex',
   },
   unemploymentRateIndigenous: {
     name: 'Indigenous unemployment rate',
     legendText: 'Indigenous',
-    format: '0[.]0%',
+    format: '0[.]0[0]%',
```
micapam: The way you've done these formats looks like a sensible heuristic. 

There's lots of repetition here, though. I reckon they should go in a constants block at the top of the file e.g.

```
FORMATS = {
  PERCENT_CONCISE: '0[.]0[0]%',
// ... etc
```

micapam (changes requested):
Needs DRYing up a bit, otherwise fine.

src/constants/indicators.js
davidgilbertson: Good idea. Fixed.

micapam (approved):
lgtm!

PR #83, opened 2017-11-21
Feat/sc 140 hide western sydney on one chart

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-140

This is - by request - the hacky, not future-proof, not logical, but quick approach to applying special rules for Western Sydney.


src/constants/indicators.js
```diff
@@ -219,6 +219,7 @@ export default {
     name: 'Jobs accessible by car within 30 minutes',
     format: FORMATS.PERCENT_ZERO_OR_ONE_DECIMAL,
     contextual: false,
+    hideForWesternSydney: true,
```
micapam: I would feel a lot better about something like: 

```
hideForCities: ['western-sydney']
```

src/components/SubCategoryCharts/SubCategoryCharts.js
```diff
@@ -9,12 +9,22 @@ import style from './SubCategoryCharts.scss';
 
 const SubCategoryCharts = (props) => {
   const subCategory = props.subCategory;
+  let charts = subCategory.charts;
 
-  const heroChart = subCategory.charts.find(chart => (
+  // SC-140 for Western Sydney, we can hide some charts
```
micapam, line 14: How about (see my note in indicators.js) - 

```
if (props.city) {
  charts = charts.filter((chart) => {
    const firstChartIndicator = INDICATORS[chart.indicatorIds[0]];
    return !firstChartIndicator.hideForCities || !firstChartIndicator.hideForCities.includes(props.city.id);
  }
}

File: src/components/SubCategorySummary/SubCategorySummary.js [hunk 3]
```diff
@@ -16,8 +16,17 @@ const SubCategorySummary = (props) => {
   const backgroundColor = getColorVariant(props.highlightColorLight);
   const cardHighlightColor = getColorVariant(props.highlightColorDark);
 
-  const indicatorCardWrappers = props.summaryIndicatorIds.map((indicatorId, i) => {
-    const lastOne = i === (props.summaryIndicatorIds.length - 1);
+  let indicatorIds = props.summaryIndicatorIds;
+
+  // SC-140 for Western Sydney, we can hide some indicator cards
```
micapam, line 21:
```
if (props.city) {
  indicatorIds = indicatorIds.filter(indicatorId => (
    !INDICATORS[indicatorId].hideForCities || !INDICATORS[indicatorId].hideForCities.includes(props.city.id)
  ));
}
```

micapam (changes requested):
I'd prefer it if we abstracted the solution slightly.

src/constants/indicators.js
davidgilbertson: The idea is that all this work will be thrown out in a week because the Western Sydney treatment will be re-done, hence the lack of generalisation.

davidgilbertson: When it's known how we will actually treat 'non cities' then we can come up with a solution to suit.

davidgilbertson: I've added https://govausites.atlassian.net/browse/SC-141 that references removing the `SC-140` code so we have no mess left behind.

micapam (approved):
Much nicer :)

PR #86, opened 2017-11-24
Improve screen-reader experience

jonalport:
https://govausites.atlassian.net/browse/SC-136

Added behaviour includes:
* Navigating between cities lands you on the overview page (as opposed to retaining subnav context) and gives focus to overview subnav item
* Navigating to subnav items gives focus to the heading pertaining to that content
* When focus is shifted, the visual outline style is removed (for that one focus only) to avoid distraction
* For links with hashes, focus is given to the element with that hash/id (Eg. for additional info pages)
* Aria labels on subnav buttons and subnav content headings, specifying which city that link/heading applies to (Eg. "Overview of Canberra", "Housing in Sydney")
* Renamed the ScrollTo... component to be more generic, as it now handles the focus stuff too


jonalport:
This also appears bigger than it is. I probably should have `git moved` those renamed files...

package.json
```diff
@@ -92,6 +92,7 @@
     "test": "node scripts/test.js --env=jsdom --colors",
     "test:coverage": "npm run test -- --coverage",
     "test:ci": "CI=true npm run test -- --coverage --runInBand --colors",
+    "update-snapshot": "jest --env=jsdom --updateSnapshot",
```
davidgilbertson: You shouldn't need this, you can just hit `u` when running tests to update the snapshots.

src/components/DoStuffOnRouteChange/DoStuffOnRouteChange.js
```diff
@@ -0,0 +1,112 @@
...
+    if (targetEl) {
+      const dims = targetEl.getBoundingClientRect();
+      targetScrollPos = dims.top + window.pageYOffset;
+    }
+  }
+
+  window.scrollTo(0, targetScrollPos);
+}
+
+class DoStuffOnRouteChange extends Component {
```
davidgilbertson: a) a less whimsical name would probably be a good idea, maybe `RouteChangeHandler`.
b) if you rename the file in a separate branch (call it tech debt) then your changes to this file will show as changes, not a whole new file. As is, all history will be lost. Not a big deal in this case.

(FYI git won't track renamed files directly, not even with `git mv`)

src/components/DoStuffOnRouteChange/DoStuffOnRouteChange.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  const outlineStyle = focusPoint.style.outline;
+
+  addOneTimeEventListener(focusPoint, 'focus', () => {
+    focusPoint.style.outline = 'none';
+  });
+
+  addOneTimeEventListener(focusPoint, 'blur', () => {
+    focusPoint.style.outline = outlineStyle;
```
davidgilbertson: This could be simplified to `focusPoint.style.outline = '';` and the `outlineStyle` variable removed. Or if you wanted to get fancy, `focusPoint.style.removeProperty('outline');`.

Setting a style property to an empty string calls `removeProperty()` under the hood anyway.

src/components/DoStuffOnRouteChange/DoStuffOnRouteChange.js
```diff
@@ -0,0 +1,112 @@
...
+
+  if (!focusPoint) {
+    return;
+  }
+
+  const outlineStyle = focusPoint.style.outline;
+
+  addOneTimeEventListener(focusPoint, 'focus', () => {
+    focusPoint.style.outline = 'none';
+  });
```
davidgilbertson: You don't need to wait until the element is focused to do this, just hide the outline and set tab index immediately, and remove them again on blur:

```js
  focusPoint.setAttribute('tabIndex', '0');
  focusPoint.style.outline = 'none';

  addOneTimeEventListener(focusPoint, 'blur', () => {
    focusPoint.removeAttribute('tabIndex');
    focusPoint.style.outline = '';
  });

  focusPoint.focus();
```

src/components/PageBanner/PageBanner.js
```diff
@@ -2,47 +2,71 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import IndicatorCard from '../IndicatorCard/IndicatorCard';
 import PageLegend from '../PageLegend/PageLegend';
-import { INDICATORS } from '../../constants';
+import { INDICATORS, ELEMENT_IDS, ALL_CITIES } from '../../constants';
```
davidgilbertson: For consistency, prefer one per line.

src/components/PageBanner/PageBanner.js
```diff
@@ -2,47 +2,71 @@ import React from 'react';
...
-    ? props.city.indicators[props.indicator]
+  const {
+    city,
+    title,
+    description,
+    colorName,
+    indicator,
+    cities,
+    isContextPage,
+  } = props;
```
davidgilbertson: I think it's better to not destructure props like this, and I should probably have that in the readme.

It makes it more difficult to discern what values are coming from props and which are declared locally, this is more apparent in bigger components where there might be half a dozen each of local variables and props.

(And just for consistency, everywhere else props are referred to by `props.xxx`).

src/components/PageBanner/PageBanner.js
```diff
@@ -2,47 +2,71 @@ import React from 'react';
...
       <div className={style.container}>
         <div className={style.textAndIndicator}>
           <div className={style.titleText}>
-            <h1 className={style.title}>{props.title}</h1>
-
-            <p className={style.description}>{props.description}</p>
+            <h1
+              id={ELEMENT_IDS.INITIAL_FOCUS_POINT_ON_PAGE}
+              className={style.title}
+              tabIndex="0"
```
davidgilbertson: I think `tabIndex` should be added programatically at the same point you're setting focus, otherwise this is always in the tab order, and can be highlighted:

image

src/components/PageBanner/PageBanner.js
```diff
@@ -2,47 +2,71 @@ import React from 'react';
...
+              aria-label={a11yText(city, title)}
```
davidgilbertson: I don't think this is necessary. Doesn't the text for the navigation link read "Infrastructure for Bendigo"? So when they click on that we don't need to repeat the city. Maybe ask Lily for a second opinion. I'd err on the side of less complexity.

src/components/PageBanner/PageBanner.js
```diff
@@ -2,47 +2,71 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import IndicatorCard from '../IndicatorCard/IndicatorCard';
 import PageLegend from '../PageLegend/PageLegend';
-import { INDICATORS } from '../../constants';
+import { INDICATORS, ELEMENT_IDS, ALL_CITIES } from '../../constants';
 import getMinAndMaxForIndicator from '../../helpers/getMinAndMaxForIndicator';
 import getColorVariant from '../../helpers/getColorVariant';
 import style from './PageBanner.scss';
 
+export const a11yText = (city, title) => {
```
davidgilbertson: You don't need to export this. The fact that it's a separate function is just an implementation detail of the component. So when testing the component you can just check the `aria-label` property. If it's right, then this function is right.

(But also, I don't think it's needed at all here, see below).

src/components/AdditionalInfoPage/AdditionalInfoPage.js
```diff
@@ -14,6 +14,7 @@ const AdditionalInfoPage = () => (
         <h1
           id={ELEMENT_IDS.ACCESSIBILITY}
           className={style.title}
+          tabIndex="0"
```
davidgilbertson: These should only be temporarily focusable. This leaves the `<h1>` in the tab order (see below).

src/components/CategoryNavContainer/CategoryNavContainer.js
```diff
@@ -1,9 +1,19 @@
...
+    return ALL_CITIES;
+  }
+
+  return cities.filter(city => city.id === id)[0].name;
+};
 
 const mapStateToProps = (state, ownProps) => ({
   categories: state.categories,
   cityId: ownProps.cityId,
+  cityName: findCityNameById(state.cities, ownProps.cityId),
```
davidgilbertson: You could probably simplify this and just pass the city if there is one, e.g. 

```js
city: state.cities.find(city => city.id === ownProps.cityId),
```

And then in the nav handle the case of 'all cities'.

src/components/CategoryNavContainer/CategoryNavContainer.js
```diff
@@ -1,9 +1,19 @@
 import { connect } from 'react-redux';
 import CategoryNav from './CategoryNav/CategoryNav';
+import { NO_CITY, ALL_CITIES } from '../../constants';
+
+export const findCityNameById = (cities, id) => {
```
davidgilbertson: Same comment as elsewhere. Rather than export implementation details for direct testing, just test that the container returns what it should.

src/components/FeedbackPage/FeedbackPage.js
```diff
@@ -1,13 +1,20 @@
...
 
 const FeedbackPage = () => (
   <PageWrapper dimFeedback>
     <div className={style.feedbackWrapper}>
       <div className={style.container}>
-        <h1 className={style.title}>Please give feedback</h1>
+        <h1
+          id={ELEMENT_IDS.INITIAL_FOCUS_POINT_ON_PAGE}
+          className={style.title}
+          tabIndex="0"
```
davidgilbertson: All these `tabIndex="0"` can go

src/constants/misc.js
```diff
@@ -2,6 +2,7 @@ export const DATA_URL = '/data/smartCitiesData.json';
 
 export const NO_CATEGORY = 'overview';
 export const NO_CITY = 'all-cities';
+export const ALL_CITIES = 'All Cities';
```
davidgilbertson: A bit further down in this file we have a `STRINGS` bucket for all strings.

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.js
```diff
@@ -42,7 +45,11 @@ const AllCitiesOverview = props => (
     <div className={style.infoWrapper}>
       <div className={style.infoContainer}>
         <div className={style.infoTextWrapper}>
-          <h1 className={style.infoTitle}>About</h1>
+          <h1
+            id={ELEMENT_IDS.INITIAL_FOCUS_POINT_ON_PAGE}
```
davidgilbertson, line 49: Love a good descriptive variable name.

davidgilbertson (changes requested):
A few things...

jonalport:
Thanks for the detailed review, @davidgilbertson! Am currently addressing.

package.json
jonalport: Cool, that's easier.

src/components/DoStuffOnRouteChange/DoStuffOnRouteChange.js
jonalport: Agreed, but how do you know it will be ''? The idea behind saving the initial value is that it restores it to whatever it was.

jonalport:
Test comment...

src/components/PageBanner/PageBanner.js
jonalport: Yeah I can see that argument, but it also makes things more concise not to have "this.props." 50 times in a file. Will take it out for consistency though 👍

src/components/PageBanner/PageBanner.js
jonalport: Yes, this is a better idea.

src/components/PageBanner/PageBanner.js
jonalport: Complexity vs consistency... Have taken it out for headings 👍

src/constants/misc.js
jonalport: Roger that.

src/components/DoStuffOnRouteChange/DoStuffOnRouteChange.js
jonalport: But where's the fun in that 😁 

I guess I assumed git mv would have eased the diffing overload but perhaps not then 😞

src/components/CategoryNavContainer/CategoryNavContainer.js
jonalport: Yep I will just pass the name down along with the ID then.

src/components/CategoryNavContainer/CategoryNavContainer.js
jonalport: Yeah I do agree with this. Black box!

src/components/DoStuffOnRouteChange/DoStuffOnRouteChange.js
davidgilbertson: Yeah I thought of that after making the comment.

src/components/DoStuffOnRouteChange/DoStuffOnRouteChange.js
davidgilbertson: Yep, they probably should never have made a `git mv` - it's just confusing. Bottom of [this page](https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository) shows that it just does a normal move.

src/components/PageBanner/PageBanner.js
davidgilbertson: You can still do `const { props } = this`.

We actually had a bug in the ICT project where none of the props are referenced with `prop.x` and someone was doing the equivalent of `cities.sort()`, not realising they were mutating state and altering sorting elsewhere.

jonalport:
I think I've now addressed all the things in the review. Is there a workflow step that I need to complete? Or @davidgilbertson do you need to confirm (to unblock)? 

😕

src/components/AdditionalInfoPage/AdditionalInfoPage.js
```diff
@@ -125,6 +127,7 @@ const AdditionalInfoPage = () => (
         <h1
           id={ELEMENT_IDS.CONTENT_CREDIT}
           className={style.title}
+
```
davidgilbertson: This is adding new blank lines, there should be no changes in this file

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
```diff
@@ -83,6 +88,7 @@ CategoryNav.propTypes = {
     id: PropTypes.string.isRequired,
   })).isRequired,
   cityId: PropTypes.string.isRequired,
+  cityName: PropTypes.string,
```
davidgilbertson: Should this be `.isRequired`?

src/components/PageBanner/PageBanner.js
```diff
@@ -2,12 +2,24 @@ import React from 'react';
...
 const PageBanner = (props) => {
+  // const {
+  //   city,
+  //   title,
+  //   description,
+  //   colorName,
+  //   indicator,
+  //   cities,
+  //   isContextPage,
+  // } = props;
```
davidgilbertson: Remove commented-out code

src/components/PageBanner/PageBanner.js
```diff
@@ -22,8 +34,11 @@ const PageBanner = (props) => {
       <div className={style.container}>
         <div className={style.textAndIndicator}>
           <div className={style.titleText}>
-            <h1 className={style.title}>{props.title}</h1>
+            <h1
+              id={ELEMENT_IDS.INITIAL_FOCUS_POINT_ON_PAGE}
+              className={style.title}
 
+            >{props.title}</h1>
```
davidgilbertson: Extra blank line can go.

src/components/PageBanner/PageBanner.test.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 import PageBanner from './PageBanner';
+import { ALL_CITIES } from '../../constants';
```
davidgilbertson: This is never used (and doesn't exist), can go.

src/components/RouteChangeHandler/RouteChangeHandler.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  if (!focusPoint) {
+    return;
+  }
+
+  const outlineStyle = focusPoint.style.outline;
+
+  addOneTimeEventListener(focusPoint, 'blur', () => {
+    focusPoint.style.outline = outlineStyle;
```
davidgilbertson: This needs to `focusPoint.removeAttribute('tabIndex')` otherwise:

image

src/components/RouteChangeHandler/RouteChangeHandler.js
```diff
@@ -0,0 +1,110 @@
...
+
+  // if we have changed cities, and landed on overview
+  // then focus on overview subnav item
+  if (!focusPoint) {
+    const oldSegments = oldPath.split('/', 3);
+    const currentSegments = currentPath.split('/', 3);
+
+    if (currentSegments[2] === NO_CATEGORY && currentSegments[1] !== oldSegments[1]) {
+      focusPoint = document.getElementById(ELEMENT_IDS.OVERVIEW_LINK);
+    }
```
davidgilbertson: A very minor, opinion thing, that isn't new in these latest changes: this might be easier to read with named variables, e.g.

```js
const oldCity = oldPath.split('/')[1];
const [, currentCity, currentCategory] = currentPath.split('/');

if (currentCategory === NO_CATEGORY && currentCity !== oldCity) {
  focusPoint = document.getElementById(ELEMENT_IDS.OVERVIEW_LINK);
}
```

src/components/RouteChangeHandler/RouteChangeHandler.test.js
```diff
@@ -0,0 +1,154 @@
...
+
+  wrapper.setProps({
+    location: {
+      pathname: '/2/1',
+    },
+  });
+
+  expect(document.getElementById).toHaveBeenCalledWith(focusElementId);
+  expect(focusOnElement).toHaveBeenCalledTimes(1);
+});
```
davidgilbertson: Since every discovered bug should have a test, this should be testing that `tabIndex` and `outline` are reset on blur.

src/components/RouteChangeHandler/RouteChangeHandler.js
```diff
@@ -0,0 +1,110 @@
...
+  let focusPoint;
+
+  // first look for a hash
+  if (document.location.hash) {
+    focusPoint = document.querySelector(document.location.hash);
+  }
+
+  // if we have changed cities, and landed on overview
+  // then focus on overview subnav item
+  if (!focusPoint) {
```
davidgilbertson: Minor (and not new in these change) - this is logically just an `else`, right? Might be easier to read like so:
```js
// first look for a hash
if (document.location.hash) {
  focusPoint = document.querySelector(document.location.hash);
} else {
  // if we have changed cities, and landed on overview
  // then focus on overview subnav item
  const oldSegments = oldPath.split('/', 3);
  const currentSegments = currentPath.split('/', 3);

  if (currentSegments[2] === NO_CATEGORY && currentSegments[1] !== oldSegments[1]) {
    focusPoint = document.getElementById(ELEMENT_IDS.OVERVIEW_LINK);
  }
}
```

davidgilbertson (changes requested):
A few issues

src/components/RouteChangeHandler/RouteChangeHandler.js
jonalport: DOH!

src/components/RouteChangeHandler/RouteChangeHandler.test.js
jonalport: Roger that

src/components/RouteChangeHandler/RouteChangeHandler.js
jonalport: Array destructure, smart. Definitely reads better.

src/components/AdditionalInfoPage/AdditionalInfoPage.js
jonalport: Ah yes, that was my find and replace, whoops. We should make the linter pick this up.

.eslintrc.js
```diff
@@ -22,6 +22,8 @@ module.exports = {
           'warn',
           'error',
         ],
-      }],
+      }
+    ],
+    'jsx-a11y/no-noninteractive-tabindex': 'off'
```
davidgilbertson: This should be reverted now since we're not breaking the rule any more.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.js
```diff
@@ -83,6 +88,7 @@ CategoryNav.propTypes = {
     id: PropTypes.string.isRequired,
   })).isRequired,
   cityId: PropTypes.string.isRequired,
+  cityName: PropTypes.string.isRequired,
```
davidgilbertson, line 91: This throws a propTypes error in the console on all cities/overview. I thought it was always passed in. So you could either put ` || STRINGS.ALL_CITIES` somewhere further up the tree or revert to not required since you have `const cityName = this.props.cityName || STRINGS.ALL_CITIES;` in this file already.

Although the preferred approach for logic like `const cityName = this.props.cityName || STRINGS.ALL_CITIES;` is to use defaultProps and keep the component cleaner. That's a minor thing though.

src/components/CategoryNavContainer/CategoryNav/CategoryNav.test.js
```diff
@@ -23,6 +23,7 @@ document.querySelector = jest.fn(() => mockElement);
 
 const defaultProps = {
   cityId: 'perth',
+  cityName: 'Perth',
```
davidgilbertson: If you revert to not having `isRequired` this can go.

src/components/RouteChangeHandler/RouteChangeHandler.test.js
```diff
@@ -0,0 +1,189 @@
...
+  wrapper.setProps({
+    location: {
+      pathname: '/2/1',
+    },
+  });
+
+  expect(document.getElementById).toHaveBeenCalledWith(focusElementId);
+  expect(element.focus).toHaveBeenCalledTimes(1);
+  expect(element.style.outline).toBe('none');
+  expect(element.setAttribute).toHaveBeenCalledWith('tabIndex', '0');
```
davidgilbertson, line 162: Nice.

src/components/RouteChangeHandler/RouteChangeHandler.test.js
```diff
@@ -0,0 +1,189 @@
...
+    { attachTo: root }
+  );
+
+  wrapper.setProps({
+    location: {
+      pathname: '/2/1',
+    },
+  });
+
+  const node = wrapper.find(`#${focusElementId}`);
```
davidgilbertson: Very minor, redundant variable, e.g. could be
```js
const domNode = wrapper.find(`#${focusElementId}`).getDOMNode();
```

src/components/RouteChangeHandler/RouteChangeHandler.test.js
```diff
@@ -0,0 +1,189 @@
...
+  // Note: trying to simulate blur event does not work
+  //   Eg) node.simulate('blur');
+  //   a it tries to call the React event handler
```
davidgilbertson: Typo? "a it tries"

davidgilbertson (changes requested):
propTypes error and some tidy up stuff

src/constants/misc.js
```diff
@@ -2,6 +2,7 @@ export const DATA_URL = '/data/smartCitiesData.json';
 
 export const NO_CATEGORY = 'overview';
 export const NO_CITY = 'all-cities';
+export const MAX_CATEGORY_NAV_TEXT_LENGTH = 14;
```
jonalport, line 5: If they are all 14, this will still be a problem. Better than nothing though.

src/constants/categories.test.js
```diff
@@ -64,6 +65,11 @@ CATEGORIES.forEach((category) => {
     expect(heroExistsInCharts(category.heroIndicatorId, category.subCategories)).toBe(true);
   });
 
+  it(`${category.navName || category.name}: name (or navName) should be ${MAX_CATEGORY_NAV_TEXT_LENGTH} chars or shorter`, () => {
+    const name = category.navName || category.name;
```
davidgilbertson, line 69: Minor. Really I just want to make one more comment...

Technically `const name = category.navName || category.name;` could go up two lines and be used inside the `it()` as well.

davidgilbertson (approved):
LGTM!

src/constants/categories.test.js
jonalport: Yes, this is true. I wanted to package it neatly together.

PR #87, opened 2017-11-27
Fix checkbox tick not showing

davidgilbertson:
Previously, ticks would only show in the first column and the first two items in the second column:
image

Amazingly, this happened regardless of the number of items in a column:
image

Now fixed:
image

The `transform-origin` bug is unrelated and hadn't been logged, but occurs in an upcoming version of Chrome and was transform-related so I slipped it in.


jonalport (approved):
This does now appear to be working. Hard to believe Chrome had a bug.

Why sun sparkles?!

davidgilbertson:
It's a lame Easter egg.

PR #88, opened 2017-11-27
Null data bands on charts

micapam:
Jira: https://govausites.atlassian.net/browse/SC-144


src/components/CityColumnChart/CityColumnChart.js
```diff
@@ -74,18 +75,42 @@ function getSeries(props) {
...
+      color: cityColor,
       from: idx - 0.5,
       to: idx + 0.5,
-    }];
+    });
   }
 
-  return [];
+  return bands;
 }
```
davidgilbertson, line 115: Super minor: it might be worth adding a comment in `convert.js` on line 38 something along the lines of:
```
// because 'value' is a string, this will include values of 0 but exclude empty cells

davidgilbertson (approved):
LGTM

PR #91, opened 2017-11-27
Make assorted copy updates

jonalport:
Bits & bobs. There are some screenshots in the Jira ticket:
https://govausites.atlassian.net/browse/SC-143


davidgilbertson:
Checking this out now...

src/constants/indicators.js
```diff
@@ -171,7 +171,7 @@ export default {
   },
   dispersion: {
     name: 'Local governments per 100k people',
-    format: FORMATS.PERCENT_ZERO_OR_ONE_DECIMAL,
+    format: FORMATS.ZERO_OR_ONE_DECIMAL,
```
davidgilbertson: I've tried to get three significant figures everywhere. So for indicators mostly under 10, two decimal places, and for indicators mostly over 10, one decimal place. So this can be `ZERO_TO_TWO_DECIMALS` or `ONE_OR_TWO_DECIMALS`. Either is fine, but if you do `ONE_OR_TWO_DECIMALS`, the y axis to the chart looks a bit neater because each label will be the same length.

image

vs

image

src/constants/indicators.js
```diff
@@ -431,15 +431,15 @@ export default {
...
+    format: FORMATS.PERCENT_ZERO_OR_ONE_DECIMAL,
     contextual: false,
     lastUpdated: '2016',
     source: 'New business entrants and exits, New Businesses as a share of the stock',
   },
   newBusinessExits: {
     name: 'Share of ceased businesses',
     legendText: 'Exits',
-    format: FORMATS.THREE_DECIMALS,
+    format: FORMATS.PERCENT_ZERO_OR_ONE_DECIMAL,
```
davidgilbertson: Minor: same comment as above, if these are forced to at least one decimal, the chart y axis looks a bit neater.

src/constants/indicators.js
```diff
@@ -431,15 +431,15 @@ export default {
   newBusinessEntrants: {
     name: 'New business entry rate',
     legendText: 'Entrants',
-    format: FORMATS.THREE_DECIMALS,
+    format: FORMATS.PERCENT_ZERO_OR_ONE_DECIMAL,
```
davidgilbertson: A question for @jamesbroadbent202 - if this is a percent, should the description be changed from "...as a share of the total..." to "...as a percentage of the total..."?

src/constants/categories.js
```diff
@@ -814,12 +824,11 @@ export const CATEGORIES = [
         ],
       },
       {
-        name: 'Digital business',
+        name: 'Digital opportunities',
```
davidgilbertson, line 826: Phew, so glad we caught that one before going live!

src/constants/categories.js
```diff
@@ -252,7 +252,7 @@ export const CATEGORIES = [
     ],
     subCategories: [
       {
-        name: 'Employment',
+        name: 'Labour Force',
```
davidgilbertson: We should be sentence case everywhere, so "Labour force".

Actually "Safety and Crisis Support" under Liveability should be fixed too.

Maybe ask Nick about the category titles "Liveability and Sustainability", "Jobs & Skills", and "Infrastructure and Investment". I would think these should all be sentence case too.

davidgilbertson:
It's not in the Jira task, but there's a typo in "Greenhouse gas emissions" -> "...which are *direcet* greenhouse emissions..."

And under "Industry sector share of employment" the axis label is "*Perecntage* of employment"

Might be fastest to just fix these in this branch. Or a new one, either way...

jonalport:
Yeah this branch works.

davidgilbertson (approved):
A few little things that would be good to fix. Or just merge.

src/constants/categories.js
jonalport: I am all for sentence case everywhere. There seems to be a lot of random capitalisations in titles, so thought that was perhaps the vibe we were going for.

jonalport:
Cool, will look into those things.

src/constants/categories.js
jonalport: What about "All Cities" - another case for sentence case I reckon?

@davidgilbertson

src/constants/indicators.js
jonalport: Yes, I like it.

src/constants/indicators.js
jonalport: Have changed to "... as a percentage of the stock" for entries & exits.

src/constants/categories.js
jonalport: Have updated that and all the others I could find.

jonalport:
Have addressed above issues and added sentence-case updates. This is ready to merge I think.

PR #92, opened 2017-11-27
Add transform-box

davidgilbertson:
This is interesting. In most current browsers, the `transform-origin` value is relative to the element it's defined upon. So the code below rotates around its own centre.

But the `transform-box` property from the [transforms spec](https://drafts.csswg.org/css-transforms/#transform-box) will be supported in Chrome in a few weeks, (and Safari sometime soon), and it defaults to `view-box`, which uses the nearest `<svg>` element. So the transform would centre around the parent svg element.

So, this sets `transform-box` to `fill-box` which replicates current behaviour.


jonalport (approved):
Huh! That is indeed interesting 💯

PR #94, opened 2017-11-29
Toggle tooltip on click

jonalport:
@davidgilbertson what do you recommend re the click event on the non-interactive element? Is there a better way to do it?


davidgilbertson:
I think it's fine. The idea behind a click event on a non-interactive element being 'bad practice' is that screen readers don't read it out as something clickable. But this text isn't accessible to screen readers anyway.

We should probably investigate using `<abbr>` for this with associated `<dl>`, but there's problems with that.

src/components/Tooltip/Tooltip.js
```diff
@@ -4,6 +4,8 @@ import classnames from 'classnames';
 import Portal from '../Portal/Portal';
 import styles from './Tooltip.scss';
 
+// const isTouch = () => typeof Touch === 'object';
+
```
davidgilbertson: Commented out code can go

src/components/Tooltip/Tooltip.js
```diff
@@ -56,13 +59,27 @@ class Tooltip extends PureComponent {
...
-      <span // a span so it's valid HTML no matter where it's used
+      // a span so it's valid HTML no matter where it's used
+      // Note:
+      //   the onClick there for touch devices
+      //   but it doesn't seem right to apply role=button
+      //   hence the eslint disabling
+      <span // eslint-disable-line
         onMouseOver={this.showTooltip}
         onMouseOut={this.hideTooltip}
+        onClick={this.toggleTooltip}
```
davidgilbertson: Minor: this has the slightly strange behaviour that if you click on desktop, the tooltip will show as soon as you hover, then hide when you click. Not a big deal, but you could instead do:

```js
onTouchStart={this.toggleTooltip}
```

davidgilbertson (changes requested):

src/components/Tooltip/Tooltip.js
davidgilbertson: FYI, if you were looking at detecting touch, there's only one way to do that that works reliably ... by waiting for the user to touch.

jonalport: Yeah decided against it, but failed to delete - whoops!

src/components/Tooltip/Tooltip.js
jonalport: Yeah I noted that but didn't really think it was a negative thing. Happy to change to that event though.

davidgilbertson (approved):
Wonderful!

PR #103, opened 2017-11-30
Feat/158 city deals

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-158

A surprisingly complex change. We had assumed until now that city descriptions would be text, but adding links makes them markup which means they can't exist in JSON.

So this leaves the JSON as just the source ID and the indicator data, and merges that with city data on load.


scripts/convert.js
```diff
@@ -5,10 +5,7 @@
 
 import csv from 'csv';
 import getStdin from 'get-stdin';
-import {
-  INDICATORS,
-  CITIES,
-} from '../src/constants';
+import INDICATORS from '../src/constants/indicators';
```
davidgilbertson, line 8: This needs to import `INDICATORS` directly from the `indicators.js` file, since the `cities.js` file now includes JSX which is invalid JavaScript (so running this convert script would break).

src/constants/cities.js
```diff
@@ -52,7 +54,7 @@ export default {
   'Greater Darwin': {
     id: 'greater-darwin',
     name: 'Darwin',
-    description: 'Darwin is a coastal city and the capital of the Northern Territory. Darwin has a tropical climate with a distinct wet and dry season and is the gateway to the World Heritage listed Kakadu National Park. Darwin spans an area of around 3,100 square kilometres.The Commonwelath and Northern Territory Government have entered into an agreement to create a City Deal for Darwin',
+    description: 'Darwin is a coastal city and the capital of the Northern Territory. Darwin has a tropical climate with a distinct wet and dry season and is the gateway to the World Heritage listed Kakadu National Park. Darwin spans an area of around 3,100 square kilometres.The Commonwealth and Northern Territory Government have entered into an agreement to create a City Deal for Darwin.',
```
davidgilbertson: Unrelated typos

src/constants/cities.js
```diff
@@ -1,8 +1,10 @@
 /* eslint-disable */
+import React from 'react';
+import Link from '../components/Link/Link';
```
micapam, line 3: I have strong reservations about this. Not just mixing concerns (view and data) but, flowing from that, the fact that the format of the cities data is no longer self-explanatory. Even someone who knows HTML would be confused by it. Why not just use Markdown (or a subset of it) to format the text?

scripts/convert.js
micapam: Seems to me that the moment at which one would be alerted to invalid data is with an error during data import - not an warning during app runtime. Plus I have reservations about use of JSX (see my comments elsewhere in this PR).

micapam (dismissed):
I'm not convinced this is the best approach, let's discuss ...

src/constants/cities.js
davidgilbertson: I have strong reservations with introducing markdown...

I get that it looks weird at first, but this is quite common practice: wherever strings are stored in a React project where those strings need to contain HTML, JSX is used.

We use this pattern already in the icon definition file which is data and JSX together, it might seem odd but it works well.

Using markdown (to mix view and data) sounds simple up front but is a messy solution for a few reasons:

1. It adds complexity - it requires a mechanism to parse the markdown to HTML then inject that into the page (the only way to do so is to use React's `dangerouslySetInnerHTML()` method), and we'd want a new component to handle this. Just for two strings.
2. To a non-developer there's a new syntax to learn, even though it's a simple one (although confusing in that for hyperlinks, Jira markdown and GitHub markdown and whatever-our-markdown is would not all be the same).
3. You can't have CSS classes with markdown so it means we can't apply the same hyperlink styling to those links without duplicating styling logic just for links that might appear in markdown.
4. We can't define a link to open externally.
4.5. And not a biggie, but it also wouldn't allow us to track analytics on outbound links if we wanted to do this through the `<Link>` component in the future.

At the end of the day markdown does the same thing as JSX, it's just an additional, more abstract syntax for mixing data and view, and instead of needing parsing at build time, it needs parsing at run time.

Thoughts? Now we have a team of three we can vote :)

scripts/convert.js
davidgilbertson: Agreed, let's sort out the JSX vs markdown thing first then look at this.

src/constants/cities.js
davidgilbertson: @jonalport your thoughts on the above?

jonalport: I have mild reservations about this, mostly from the perspective of making non-developers confused whilst they edit this content. If these files are going to be updated by a small set of people on an infrequent basis, then maybe it's ok. But it does seem as though something like MD might be more suited.

https://www.npmjs.com/package/markdown-to-jsx 

🤔

jonalport: My go-to quick solution probably would have been some kind of custom token that allows the insertion of a link. Something like:

`((Google|http://google.com/))`

You can still style links in the same way by targeting the parent. Alternatively, you could also have your parsing function/component/library adding whatever classes you like. Also who opens in external windows, really? 😄

micapam: Wouldn't standard markdown syntax be just as easy to parse - plus we aren't making up a format ? 

`[Google](http://google.com/)`

We wouldn't even have to support full markdown, just the link syntax. E.g. a pattern like:

`/\[(.*?)\]\((.*?)\)/`

jonalport: Yep, or that.

davidgilbertson: Either way, we're writing a parser that creates HTML out of a string. We already have a thing that parses strings to HTML: JSX. It's what the entire site is written using :)

More things for the list: it's an increased surface area for bugs (on last day of dev) and poor security (the method is named `dangerouslySetInnerHTML` for good reason).

What's the payoff?

davidgilbertson: Thinking about markdown, this is what we would require. This is a crazy amount of complexity for no benefit if you ask me:
* new library to parse markdown (if we can live without anchor tags opening in a new tab) OR  a regex that only does `<a>` tags and hard codes `target="_blank"` and a _lot_ of unit tests. Because regex.
* new component that injects HTML into the DOM with `dangerouslySetInnerHTML`
* preferably a library to sanitise that data to prevent against XSS attacks
* new Sass mixin to abstract the existing link styles out of the existing `<Link>` component
* new CSS to apply that link mixin to the existing link and to the very specific case of any links that appear in this one text field used for the page banner description.

micapam: Given that all options are nasty in one way or another, we _could_ just have raw HTML in there ...

jonalport: Yeah, sounds like it would be a lot simpler to do MD if we had a generated/server-rendered HTML site... 😁 

For the record; I don't think you should change it, just registering my mild reservation.

It's true that the whole site is written in JSX but the people editing the content are not necessarily the same as the ones who wrote the site.

micapam (commented):
Because I don't have a clear suggestion that would be pragmatic esp. at this late stage, I'm changing my change request to just a comment ...

PR #110, opened 2017-12-04
Feat/sc 30 analytics

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-30

There's two parts to this
1. capturing the items in 2d of [this doc](https://docs.google.com/document/d/1jjYPrYkb9s1ziouiWIGEI5ozbfP5SOOsywEiykMzFqQ/edit). Most of them just work, but we weren't capturing individual page views.
2. capturing performance metrics. Via GA isn't the best way to do this, but now we can say we did.

To test, in the network panel, filter by `collect` and check the requests being sent when you change pages.
image

You can also log into GA if you have access and check the realtime stats.


public/index.html
```diff
@@ -29,24 +29,21 @@
     <meta property="og:description" content="Australian Government dashboard to measure the performance of Australia's largest cities against the National Cities Performance Framework.">
     <meta property="og:title" content="Smart Cities Performance Dashboard">
     <meta property="og:url" content="https://smart-cities.dashboard.gov.au">
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'UA-61222473-12');
+    </script>
```
davidgilbertson, line 37: Moved this up here to ensure `gtag` exists regardless of where the react script gets injected below.

public/index.html
```diff
@@ -29,24 +29,21 @@
...
   </head>
   <body>
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
-    <script>
-      window.__DATAVIZKIT_CONFIG__ = {
-        accessibility_module: false,
-      }
-    </script>
```
davidgilbertson, line 41: Unrelated junk removal

src/components/RouteChangeHandler/RouteChangeHandler.js
```diff
@@ -81,7 +82,17 @@ class RouteChangeHandler extends Component {
 
   componentDidUpdate(prevProps) {
     // when changing from one page to another
-    if (this.props.location !== prevProps.location) {
+    if (this.props.location.pathname !== prevProps.location.pathname) {
```
davidgilbertson, line 85: This was actually a hidden bug. The location object is new (with strict equals) even when the pathname doesn't change.

jonalport:
Are we only tracking in production env? Or should we only be, if not?

Also, have you seen this thing:
https://github.com/googleanalytics/autotrack

jonalport (approved):

davidgilbertson:
We've got a filter (I believe, @micapam knows about it) that will exclude hits coming from localhost.

PR #112, opened 2017-12-04
 Update all cities intro & context

jonalport:
https://govausites.atlassian.net/browse/SC-152

image

image


src/components/CategoryOverview/CategoryOverview.js
```diff
@@ -51,6 +53,15 @@ const CategoryOverview = (props) => {
...
     >
       <div className={style.category}>
+        {isContextCategory && (
+          <div className={style.categoryIcon}>
+            <Icon
+              icon="city"
+              color={COLORS.GREY_700}
+            />
+          </div>
+        )}
```
davidgilbertson: You can simplify this to just always show an icon, and dynamically set the color. In `categories.js`, add `iconId: 'city',` to line 19 (that defines the icon for the context category).

Then in this file:
```jsx
<Icon
  icon={props.category.iconId}
  color={categoryArrowColor}
/>
```

And probably rename `categoryArrowColor`

src/components/CategoryOverview/CategoryOverview.js
```diff
@@ -37,6 +37,8 @@ const CategoryOverview = (props) => {
       color: COLORS.WHITE,
     };
 
+  const categoryArrowColor = isContextCategory ? COLORS.GREY_700 : categoryDarkColor;
```
davidgilbertson: This (and the icon) are `GREY_600` in the design (`#757575`) in InVision. Do you know how to get the colours from InVision with "inspect mode"?

src/components/CategoryOverview/CategoryOverview.js
```diff
@@ -65,20 +76,30 @@ const CategoryOverview = (props) => {
...
 
               <p className={style.subCategoryWrapperText}>{STRINGS.SUB_CATS_INCLUDE}</p>
 
               <div>
                 {props.category.subCategories.map((subCategory) => {
                   const url = `/${cityUrlPart}/${props.category.id}#${getSubCategorySectionId(subCategory.name)}`;
 
+                  const pillStyle = isContextCategory ? {
+                    background: COLORS[subCategory.highlightColorLight],
+                    border: `2px solid ${COLORS[subCategory.highlightColorDark]}`,
```
davidgilbertson: Minor, Pills always have a 2px border, so this could just be `borderColor: COLORS[subCategory.highlightColorDark],`

Less minor: if you use `getColorVariant(subCategory.highlightColorDark)` instead of `COLORS[subCategory.highlightColorDark]` the `getColorVariant()` function will warn if the colour is not known. Otherwise it's a pretty brittle link between a text value in `categories.js` and a matching colour name being exported from `_colors.scss`.

src/constants/categories.js
```diff
@@ -17,9 +17,11 @@ export const CATEGORIES = [
     id: CATEGORY_IDS.CONTEXT,
     colorName: COLOR_NAMES.PRIMARY,
     navName: 'Context',
-    name: 'Cities context',
-    description: 'Contextual indicators can help to understand why a city performs the way it does and what policies may be effective for improving economic performance and quality of life.',
-    shortDescription: 'Contextual indicators can help to understand why a city performs the way it does and what policies may be effective for improving economic performance and quality of life.',
+    name: 'Context',
+    description: 'Context indicators show the characteristics of a city that help us understand what it’s like & why it functions the way it does.',
+    shortDescription: 'Context indicators show the characteristics of a city that help us understand what it’s like & why it functions the way it does.',
+    allCitiesDescription: 'A view of all context indicators to demonstrate what the cities are like & how they function.',
```
davidgilbertson: Is `allCitiesDescription` ever used?

src/constants/misc.js
```diff
@@ -51,7 +51,7 @@ export const LINKS = {
 };
 
 export const STRINGS = {
-  SUB_CATS_INCLUDE: 'Sub categories include:',
+  SUB_CATS_INCLUDE: 'Categories include:',
```
davidgilbertson: `SUB_CATS_INCLUDE` should probably be `CATS_INCLUDE` now.

src/style/_misc.scss
```diff
@@ -20,7 +20,18 @@ $z-index-tooltip: 5;
   }
 }
 
-// used primarily for hiding a11y text,
+// Applies a small inset shadow
+@mixin shadow-bottom {
+  box-shadow: inset 0 -3px 7px $color-black-10;
```
davidgilbertson: Oh I always thought that shadow was a mistake in the design.

I would either do `$color-black-10` inline as `rgba(0, 0, 0, 0.1)`, OR move all the other inline `rgba` colours in this file out to `_colors.scss`.

I'd worked with the idea that these are tightly bound to the shadow they're used with so don't need to be abstracted.

Also, minor, `@shadow-inset`?

src/style/_misc.scss
```diff
@@ -20,7 +20,18 @@ $z-index-tooltip: 5;
...
+}
+
+@mixin unstyled-list {
```
davidgilbertson: Is this ever used? Can go I think.

src/style/_typography.scss
```diff
@@ -0,0 +1,11 @@
...
+$font-size--default: 22px;
+
+// Font weights
+$font-weight--lighter: 200;
+$font-weight--default: 300;
+$font-weight--heavier: 500;
+$font-weight--heaviest: 700;
+
+// Line heights
+$line-height--default: 32px;
```
davidgilbertson: Oh you should have seen our enormous typography setup and been here for the hours of discussions about the best approach. The end result is that we tore it out. https://github.com/govau/smart-cities/pull/45/files#diff-e8329234d00e24af5e88be231786137f

The reason is that everything is so bespoke that there is no often-used anything. There is no 'default' font size or a 'default' weight. There might be one font size that is used more than any other (I believe that's 14px) but there's no benefit to singling that one font-size out.

And having `$font-weight--heavier` just adds a layer of complexity. We're never going to change `$font-weight--heavier` to some other value in bulk, so why type `$font-weight--default` when I could type `300`? Same with `22px`, we would never change that en masse, so no point having it as a variable.

In the style guide at `/style-guide` you can see the 18 different combinations that are currently used, and it's grown by two or three since that was done about four weeks ago. As much as I'd love to have a rock solid set of types, it just doesn't work with this site because there's too many of them.

I think this file can go.

src/components/PageLegend/PageLegend.scss
```diff
@@ -0,0 +1,110 @@
+@import '~style/main';
+
```
davidgilbertson: This file should be deleted (it was removed in another branch)

src/components/CategoryOverview/CategoryOverview.scss
```diff
@@ -72,12 +73,6 @@
   line-height: 1;
   margin-bottom: 16px;
 
-  .category__context & {
```
davidgilbertson, line 75: `.category__context` is used in one other place here and I'm pretty sure it can go. It then doesn't need to be set in `CategoryOverview.js` which means `classnames` can be removed and the reference to `category__context` in the test can go too.

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
```diff
@@ -123,39 +123,77 @@ $round-image-width: 220px;
...
-  display: flex;
-  flex-direction: column;
   padding-top: 42px;
   padding-bottom: 42px;
+}
 
-  @include for-tablet-portrait-up {
-    flex-direction: row;
+.legend {
+  margin: 40px 0 12px;
```
davidgilbertson: Try and avoid setting margin-top anywhere, then enjoy never having to think about collapsing margins again. Instead set margin-bottom on `.subInfoBodyText`

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.js
```diff
@@ -82,24 +82,47 @@ const AllCitiesOverview = props => (
...
-      </div>
 
-      <Icon
-        className={style.subInfoIcon}
-        icon="city"
-        size={120}
-        color={COLORS.PRIMARY_600}
-      />
+        <dl className={style.legend}>
+          <dt className={style.term}>
```
davidgilbertson: I feel this would be cleaner and simpler CSS if you wrapped each row in a `<div>`

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
```diff
@@ -123,39 +123,77 @@ $round-image-width: 220px;
...
-  margin-bottom: 32px;
+.term {
+  font-weight: $font-weight--heavier;
+  margin: 0 0 10px;
 
-  @include for-tablet-portrait-up {
-    margin-bottom: 0;
-    margin-right: 64px;
+  @include for-tablet-landscape-up {
+    flex: 1 0 26%;
```
davidgilbertson: You should generally never need to set basis directly (the spec recommends against it). Instead just set the width. That way you can just define the width of the title (with `flex: none` set which is nice and obvious) and have the description fill the rest (with `flex: auto`) so you don't need to set the width manually for these.

Also, since you essentially want a consistent gap between the right end of "Performance" and the description, you can simplify even further by just setting the width and you don't even need the breakpoints. So:

```scss

.term {
  font-weight: $font-weight--heavier;
  margin: 0 0 10px;
  flex: none;

  @include for-tablet-landscape-up {
    width: 200px;
    margin-bottom: 20px;

    &:nth-last-child(2) {
      margin-bottom: 0;
    }
  }
}

.definition {
  margin: 0 0 30px;
  flex: auto;

  &:last-child {
    margin-bottom: 0;
  }

  @include for-tablet-landscape-up {
    margin-bottom: 20px;
  }
}
```

And if you wrap each row in a `<div>` you can do away with the clunky `&:nth-last-child(2)` styles.

src/components/CategoryOverview/CategoryOverview.js
```diff
@@ -65,20 +76,30 @@ const CategoryOverview = (props) => {
...
+                  } : null;
+
                   return (
                     <NavLink
                       key={subCategory.name}
                       to={url}
                     >
-                      <Pill className={style.subCategoryLink} shadow>
+                      <Pill className={style.subCategoryLink} style={pillStyle} shadow>
```
davidgilbertson: In the designs, the shadows are the opposite way around (between the sub-categories and the link on the right).

This branch:
image

Design:

image

src/components/CategoryOverview/CategoryOverview.test.js
```diff
@@ -68,8 +68,8 @@ describe('when rendering the "context" category', () => {
     expect(component.find('.categoryWrapper').hasClass('category__context')).toBe(true);
   });
 
-  it('should not render a category icon', () => {
-    expect(component.find('.categoryIcon').length).toBe(0);
+  it('should render a category icon', () => {
+    expect(component.find('.categoryIcon').length).toBe(1);
```
davidgilbertson: This test can actually go since it's no longer conditional on the context category (there's a few other tests in here that are no longer conditional that can be moved/removed)

davidgilbertson (changes requested):
A few things, back to you.

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.js
jonalport: Indeed! I was under the impression it wasn't valid to do that, but appears it is!

davidgilbertson: I had to Google it first, but yep, valid.

src/components/CategoryOverview/CategoryOverview.js
jonalport: Hah yeah I know about inspect, I just used an existing colour which I thought was pret-ty close.

src/style/_typography.scss
jonalport: What if we want to change the font someday and 300 is not a supported weight? Or what if we wanted to proportionally increase the font size across the site? Seems to me that having a DRY variable approach would be optimal.

jonalport:
Updated!

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
```diff
@@ -123,36 +123,57 @@ $round-image-width: 220px;
...
+  }
+
+  @include for-tablet-landscape-up {
+    display: flex;
+    margin-bottom: 10px;
+  }
+}
+
+.icon {
+  margin-right: 14px;
```
davidgilbertson: This icon doesn't quite align correctly. Its vertical-align is middle but the other text will default to baseline. You can set
```css
vertical-align: -8px;
```

To align it nicely.

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
```diff
@@ -123,36 +123,57 @@ $round-image-width: 220px;
...
-  margin: 0 auto;
+.definition {
+  margin: 0;
 
-  @include for-tablet-portrait-up {
-    flex: none;
-    width: 140px;
-    margin: 14px 0 0; // visually align the top with the title text
+  @include for-tablet-landscape-up {
+    flex-grow: 1;
```
davidgilbertson: This isn't needed, is it? This text will just expand to take up whatever space it needs. If it is needed, `flex: auto` should be the preferred method for saying 'take all available space'.

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
```diff
@@ -123,36 +123,57 @@ $round-image-width: 220px;
...
+.icon {
+  margin-right: 14px;
+}
+
+.term {
+  font-weight: 500;
+  margin-bottom: 10px;
+
+  @include for-tablet-landscape-up {
+    flex-shrink: 0;
```
davidgilbertson: I think for the benefit of those less familiar with flexbox, `flex: none;` is much clearer. You should almost never need to set flex-shrink (or basis, or grow) directly (spec recommends using the shorthand syntax instead). Do you know about [the shorthands](https://www.w3.org/TR/css-flexbox-1/#flex-common) for `flex`? They should cover everything.

davidgilbertson (changes requested):
Just some minor things, but they add up to slightly-more-than-minor.

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
jonalport: Can also set to 100% high to avoid a hardcoded pixel value.

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
jonalport: I sometimes use shorthand in the form `flex: x x x` which I thought set shrink, grow and basis without explicitly saying so, which I had thought was actually *more* confusing than listing them each out individually.

src/components/AllCitiesOverviewContainer/AllCitiesOverview/AllCitiesOverview.scss
jonalport: I did actually notice that alignment thing previously but forgot to fix it, whoops!

davidgilbertson (approved):

jonalport:
Only 28 comments! 💯

PR #118, opened 2017-12-06
Update data to Dec 6

davidgilbertson:
Jira: https://govausites.atlassian.net/browse/SC-171

The primary change was new LinkedIn Data, but I noticed formatting in the Excel file/Google sheet was masking precision when the thing was being exported to CSV, so that's been fixed.


scripts/convert.js
```diff
@@ -16,6 +16,8 @@ const indicators = Object.entries(INDICATORS).map(([key, indicator]) => ({
 getStdin().then((rawCsv) => {
   csv.parse(rawCsv, { columns: true }, (err, data) => {
     const cities = data.map((row) => {
+      if (!row.Cities) return null;
```
micapam, line 19: I guess this (accompanied by the `filter` call below) excludes any rows without a city name, right? Is that a scenario we've encountered?

public/data/smartCitiesData.json
```diff
@@ -119,7 +119,7 @@
       "peakDelay": 0.35,
       "accessToPublicTransport": 0.112,
       "grossParklandArea": 0.693,
-      "greenhouseGasEmissions1": 4,
+      "greenhouseGasEmissions1": 3.582318687,
```
micapam, line 122: That spreadsheet behaviour is strange. I guess in the spreadsheet view of the world, whatever underlying data there is, export to CSV should match what you see on the screen as closely as possible...?

micapam (approved):
Approved.

scripts/convert.js
davidgilbertson: Yeah I encountered it this time around. I don't know if that's because this time I went via Excel instead of straight out of Google sheets.

PR #119, opened 2017-12-06
Update sidebar styles

jonalport:
image


src/constants/cities.js
```diff
@@ -38,7 +38,7 @@ export default {
   },
   'Gold Coast - Tweed Heads': {
     id: 'gold-coast-tweed-heads',
-    name: 'Gold Coast-Tweed',
+    name: 'Gold Coast-Tweed Heads',
```
davidgilbertson, line 41: Maybe make sure that whoever asked for this understands that it affects the labels on the chart axis too. We've gone backwards and forwards on this a few times.

davidgilbertson (approved):
Wonderful!

src/constants/cities.js
jonalport: Yeah Nick was aware of the chart thing too.

PR #121, opened 2017-12-07
Hotfix/remove basic auth

micapam:
Hotfix, based on previous release commit, to be applied tonight. Can be merged to master any time


jonalport (dismissed):
OG image is in here too?

PR #122, opened 2017-12-07
Updates to context colours & copy

jonalport:
Now with more purps.

https://govausites.atlassian.net/browse/SC-174
https://govausites.atlassian.net/browse/SC-170

All cities overview:
image

All cities context:
image

Specific city overview:
image

Specific city context:
image

More content in "context" & "jobs & skills":
image


src/constants/categories.js
```diff
@@ -19,9 +20,9 @@ export const CATEGORIES = [
     iconId: 'city',
     navName: 'Context',
     name: 'Context',
-    description: 'Context indicators show the characteristics of a city that help us understand what it’s like & why it functions the way it does.',
-    shortDescription: 'Context indicators show the characteristics of a city that help us understand what it’s like & why it functions the way it does.',
-    allCitiesShortDescription: 'A view of all context indicators to demonstrate what the cities are like & how they function.',
+    description: 'Context indicators show the characteristics of a city that help us understand what it\'s like and why it functions the way it does.',
+    shortDescription: <span>Characteristics of a city that help us understand what it&apos;s like and why it functions the way it does, based on the measurement framework. <br /><br />This section is a collection of Context indicators.</span>,
+    allCitiesShortDescription: <span>Characteristics of a city that help us understand what it&apos;s like and why it functions the way it does, based on the measurement framework. <br /><br />This section is a collection of Context indicators.</span>,
```
davidgilbertson: Why replace the `’` curly apostrophes (that are just normal UTF-8 characters) with straight quotes (which are HTML entities) then escape them? They should match the rest of the site (which uses curly ones)

davidgilbertson: "This section is a collection of Context indicators" - aaaah, it all makes sense now.

davidgilbertson: That should be lower case 'c' for context.

src/helpers/getColorOfHeroIndicator.js
```diff
@@ -0,0 +1,13 @@
...
+  category.subCategories.forEach((subCategory) => {
+    subCategory.charts.forEach((chart) => {
+      if (chart.indicatorIds.includes(category.heroIndicatorId)) {
+        colorName = subCategory.colorName;
+      }
+    });
+  });
+
+  return colorName;
+}
```
davidgilbertson: Instead of creating a new function that returns the colour name, you should modify the existing `getSubCategoryOfHeroIndicator` to return the whole subCategory (instead of just the name property).

This file can then go.

src/components/AllCitiesCategoryContainer/AllCitiesCategory/AllCitiesCategory.js
```diff
@@ -14,6 +15,7 @@ const AllCitiesCategory = props => (
       description={props.category.description}
       indicatorId={props.category.heroIndicatorId}
       indicatorSubCategory={getSubCategoryOfHeroIndicator(props.category)}
+      indicatorColorName={getColorOfHeroIndicator(props.category)}
```
davidgilbertson: My `indicatorSubCategory` was poorly named here, since really it's the `indicatorSubCategoryName`. So, if you leave the prop name but change `getSubCategoryOfHeroIndicator` to return the whole subCategory, rather than just the name, then you will get the colour down in `<PageBanner>` automatically.

(You will need to change the propType of `indicatorSubCategory` from string to shape with a `name` and a `colorName`.)

src/components/CityCategoryContainer/CityCategory/CityCategory.js
```diff
@@ -23,6 +24,7 @@ const CityCategory = props => (
       cities={props.cities}
       city={props.city}
       indicatorSubCategory={getSubCategoryOfHeroIndicator(props.category)}
+      indicatorColorName={getColorOfHeroIndicator(props.category)}
```
davidgilbertson: Same as above. If `getSubCategoryOfHeroIndicator` returned the whole sub-category, this file would go unchanged.

src/constants/misc.js
```diff
@@ -4,6 +4,7 @@ export const NO_CATEGORY = 'overview';
 export const NO_CITY = 'all-cities';
 export const HERO_INDICATOR_ID = 'population';
 export const HERO_INDICATOR_CATEGORY = 'Planning';
+export const HERO_INDICATOR_COLOR_NAME = 'planning';
```
davidgilbertson, line 7: I would say this is repetition, but I think the whole colour thing is a mess so...

src/constants/misc.js
```diff
@@ -56,7 +57,7 @@ export const LINKS = {
 
 export const STRINGS = {
   CATEGORIES_INCLUDE: 'Categories include:',
-  CONTEXTUAL_DEFINITION: 'Contextual indicators highlight the characteristics of a city that are not amenable to or potentially appropriate for local policy intervention but they can help to explain why a city performs the way it does.',
+  CONTEXTUAL_DEFINITION: 'Context indicators show the characteristics of a city that help us understand what it’s like & why it functions the way it does, based on the measurement framework.',
```
davidgilbertson: Minor: for consistency, '&' should be 'and'.

Also, it's a content thing, but I feel the need to point out: "the measurement framework" - is that supposed to mean something to the user?

davidgilbertson (changes requested):

src/components/AllCitiesCategoryContainer/AllCitiesCategory/AllCitiesCategory.js
jonalport: Ok that works

src/constants/categories.js
jonalport: I dunno, I guess I just remove curlies out of habit... That is a good point.

src/constants/misc.js
jonalport: I doubt it, but so much doesn't make sense.

jonalport:
@davidgilbertson - over to you.

src/components/CategoryOverview/CategoryOverview.js
```diff
@@ -39,6 +40,14 @@ const CategoryOverview = (props) => {
   const cityUrlPart = props.city ? props.city.id : NO_CITY;
   const categoryUrl = `/${cityUrlPart}/${props.category.id}`;
 
+  const indicatorCardColorName = isContextCategory
+    ? getSubCategoryOfHeroIndicator(props.category).colorName
```
davidgilbertson, line 44: Minor: this is a bit brittle. if `getSubCategoryOfHeroIndicator()` returns null, this will throw. So really `getSubCategoryOfHeroIndicator` should be throwing (or at least warning) if no sub category is found.

This is already covered by a unit test (hence minor), but it would be good to show in that function that returning `null` is not a valid thing to be doing.

src/components/CategoryOverview/CategoryOverview.test.js
```diff
@@ -7,6 +7,8 @@ import { CATEGORY_IDS } from '../../constants';
 jest.mock('../Icon/Icon', () => 'Icon');
 jest.mock('../IndicatorCard/IndicatorCard', () => 'IndicatorCard');
 jest.mock('../Pill/Pill', () => 'Pill');
+jest.mock('../../helpers/getSubCategoryOfHeroIndicator', () => () => ({ name: 'mock sub category', colorName: 'primary' }));
+
```
davidgilbertson: Minor: shouldn't have two blank lines

src/components/PageBanner/PageBanner.js
```diff
@@ -42,15 +42,15 @@ const PageBanner = props => (
...
+              <span> {props.indicatorSubCategory.name}</span>
             </div>
           )}
 
           <IndicatorCard
             size="large"
             className={style.card}
-            color={getColorVariant(props.colorName, '500')}
-            colorName={props.colorName}
+            color={props.indicatorSubCategory && props.indicatorSubCategory.colorName ? getColorVariant(props.indicatorSubCategory.colorName, '500') : getColorVariant(props.colorName, '500')}
```
davidgilbertson: How is this not causing an ESLint max-len error?

This needs to be broken out into a variable for readability (and line length) - so the whole component now needs an explicit return. Also, there's no point in checking for `props.indicatorSubCategory` and `props.indicatorSubCategory.colorName` here since the app would have broken elsewhere (and also for the `colorName` prop below) if there wasn't a `colorName`.

We should expect that if `indicatorSubCategory` is will always have those props (they should both be `.isRequired` in prop types below).

src/helpers/getSubCategoryOfHeroIndicator.js
```diff
@@ -1,13 +1,13 @@
 export default function getSubCategoryOfHeroIndicator(category) {
-  let subCategoryName = null;
+  let subCategory = null;
```
davidgilbertson, line 2: Mentioned above, returning `null` should not be valid, should warn or throw an error in that case (assuming it's covered by tests)

davidgilbertson (changes requested):
A couple of little things

src/components/PageBanner/PageBanner.js
jonalport: "is..."?

This is probably due to context categories having nested categories, but colorName is not present on the subcategory object most of the time (Eg. in "Labour force") so we fall back to the colorName prop. Looks like the indicatorSubCategory check can be removed though 👍

src/components/CityOverviewContainer/CityOverview/CityOverview.js
```diff
@@ -6,6 +6,7 @@ import PageWrapper from '../../PageWrapper/PageWrapper';
 import {
   HERO_INDICATOR_ID,
   HERO_INDICATOR_CATEGORY,
+  // HERO_INDICATOR_COLOR_NAME,
```
davidgilbertson: What's happened here? Is `HERO_INDICATOR_COLOR_NAME` is no longer used? It's now hard coded as `infrastructure` but that's not right is it? Shouldn't it be planning?

src/components/PageBanner/PageBanner.js
```diff
@@ -8,58 +8,64 @@ import {
...
-              <span className={style.heroSubCategoryText}>
-                Key indicator
-              </span>
+                <span> Showing range</span>
+              </div>
+            )}
+          </div>
+
+          <div>
+            {props.indicatorSubCategory && (
```
davidgilbertson: `{props.indicatorSubCategory && (` is redundant isn't it? This is required and will always be true, so can go.

davidgilbertson (changes requested):
A few little things

src/components/CityOverviewContainer/CityOverview/CityOverview.js
jonalport: Whoops, I was testing something and forgot to change back. Well spotted.

src/components/PageBanner/PageBanner.js
jonalport: This is true

davidgilbertson (approved):
LGTM

PR #124, opened 2017-12-12
Add raw data link to footer

jonalport:
https://govausites.atlassian.net/browse/SC-87

Added in the the raw data footer link and tidied up the markup a little.

Copyright text colour updated to be a bit darker and still within contrast guidelines.

"Privacy policy" was splitting across two lines, so fixed that too.

New:

image

image

Old:

image


src/components/FooterContainer/Footer/Footer.js
```diff
@@ -77,23 +77,31 @@ const Footer = props => (
...
+            {': '}
+            <a
+              href={`mailto:${EMAIL_ADDRESSES.CITY_PERFORMANCE}?subject=${STRINGS.CONTACT_SUBJECT}`}
+            >{EMAIL_ADDRESSES.CITY_PERFORMANCE}</a>
+          </div>
+
+          <div className={style.footerLinksRow}>
+            <a
+              href={LINKS.RAW_DATA}
+              target="_blank"
```
davidgilbertson: Any link with `target="_blank"` must set `rel="noopener noreferrer"` to protect against [this vulnerability](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/) (however unlikely that is).

We've got a `<Link>` component that does this for you, but it also adds styling, so either adding `rel` or using `<Link>` is fine.

src/constants/misc.js
```diff
@@ -67,4 +68,5 @@ export const STRINGS = {
   ALL_CITIES: 'All cities',
   CONTACT: 'Got a question on the Performance Framework? Contact PM&C',
   CONTACT_SUBJECT: 'Performance Framework',
+  RAW_DATA: 'View all raw data at dta.gov.au',
```
davidgilbertson: Should be `data.gov.au`, no?

davidgilbertson (changes requested):
Two things

src/constants/misc.js
jonalport: Yes that is probably right. I was following the designs! Also DTA...

src/components/Link/Link.js
```diff
@@ -3,21 +3,35 @@ import PropTypes from 'prop-types';
...
+      rel="noopener noreferrer"
+      target="_blank"
+    >
+      {props.children || props.href}
+    </a>
+  );
+};
 
 Link.propTypes = {
+  unstyled: PropTypes.bool,
```
davidgilbertson, line 27: Nice.

src/components/Link/Link.js
```diff
@@ -3,21 +3,35 @@ import PropTypes from 'prop-types';
...
   children: PropTypes.string,
   className: PropTypes.string,
   href: PropTypes.string.isRequired,
 };
 
+Link.defaultProps = {
+  unstyled: false,
```
davidgilbertson, line 34: Defaulting a prop to something falsey is redundant, but I'll allow it. (Unless you want to remove it and add a style section to the readme and put this in there as our very first rule.)

davidgilbertson (approved):
LGTM, with one optional thing.

PR #126, opened 2017-12-21
Remove minister photo

jonalport:
The new hotness:

image


davidgilbertson (dismissed):
Whenever I approve something without checking it first, there is almost always a bug.

jonalport:
Still good:

image

davidgilbertson (approved):
Now that's something I should really check out and check...

